PR #283, opened 2015-10-07
(EZ-56) Add reload to all ezbake service scripts

haus:
Previously the init scripts for ezbake projects did not include reload
because trapperkeeper did not support reloading applications (TK-202).
As work on that issue is underway, this commit adds reload everywhere
using SIGHUP. Anywhere that had force-reload defined now calls reload
instead of restart.

### DIFF
--- a/resources/puppetlabs/lein-ezbake/template/foss/ext/debian/ezbake.init.erb
+++ b/resources/puppetlabs/lein-ezbake/template/foss/ext/debian/ezbake.init.erb
@@ -197,4 +197,4 @@ case "$1" in
         ;;
 esac
 
-:
+exit 0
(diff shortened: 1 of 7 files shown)


haus:
@cprice404 @pcarlisle if you get a PR open/merged, just shout here and we can get this in.

cprice404:
yay, thanks @haus , not sure when our other thing will make it into a formal PR but I'll make sure to ping here when it does.

stahnma:
Still blocked....

cprice404:
For posterity, this is probably blocked (and somewhat replaced) by the epic SERVER-1237.

stahnma:
Is this needed now that #323 was merged?

haus:
@stahnma a version of it is still required yes.

haus:
the new version will simply call the reload command instead of what this PR does

PR #285, opened 2015-10-23
(PDB-2058) Modify project.clj for publishing to clojars

kbarber:
Since we are now public, we should be pushing to clojars instead
of our internal repository.

Signed-off-by: Ken Barber <ken@bob.sh>

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,12 +1,9 @@
-(defn deploy-info
-  [url]
-  { :url url
-    :username :env/nexus_jenkins_username
-    :password :env/nexus_jenkins_password
-    :sign-releases false })
-
 (defproject puppetlabs/lein-ezbake "0.2.11-SNAPSHOT"
   :description "A system for building packages for trapperkeeper-based applications"
+  :url "https://github.com/puppetlabs/ezbake"
+  :license {:name "Apache License 2.0"
+            :url "http://www.apache.org/licenses/LICENSE-2.0"}
+
   :dependencies [[me.raynes/fs "1.4.6" :exclusions [org.clojure/clojure]]
                  [me.raynes/conch "0.8.0"]
                  [clj-time "0.6.0"]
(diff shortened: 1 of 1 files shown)


kbarber:
@haus @waynr @melissa fyi ^^. Happy to debate this, I'm pretty sure there are build considerations that need to be thought-through.

haus:
@kbarber this looks reasonable to me. i'm still interested in how we handle tk projects (do we push to both or just to clojars?).

kbarber:
Seems to me we just push to clojars, and rely on the mirrors locally. Sorry /cc @cprice404 also.

I'm wary of any cases where we might be using snapshots and such for other groups and purposes, I'm hoping by now we're using a local repo technique instead of round-robining via nexus ... this is why we should wait for @waynr to weigh in on this, since he did most of that work.

haus:
sounds good to me.

waynr:
@kbarber there are no CI jobs that are deploying snapshot versions of `lein-ezbake`--the jobs I am aware of that deploy `lein-ezbake` only deploy release versions. 

Also, did I understand you correctly this morning when you asked that we figure out a way to push all of our internal `lein-ezbake` nexus artifacts to clojars? I don't know what the best way to do that is, would you mind filing a QENG ticket with this request so we can put some effort into figuring it out and documenting for similar needs in the future?

waynr:
Oh yeah, :+1:

kbarber:
@waynr In regards to older jars, I was simply presenting the idea of pushing some (not all) old jars to clojars that we have for shipped PDB instances so our source based installations for those revs will work today. I'm not sure I need to engage you or QENG for this task in particular yet, but I'll let you know.

kbarber:
I've updated this PR to include more metadata that lein deploy requires, else it complains. Also without the git scm setup, it tries to tell clojars to use my repository on the website, which is not valid :-).

waynr:
@kbarber okay, gotcha. I'm :+1: on this, are you ready to merge?

kbarber:
@waynr yeah sure, why not :-).

PR #286, opened 2015-10-29
(EZ-57) Attempt to pull ezbake.conf from both project and upstream jars

camlow325:
In previous commits, ezbake would only try to pull an ezbake.conf file
from upstream dependencies of the jar being packaged.  This commit changes
the logic to also pull in an ezbake.conf file that may be embedded in
the immediate project jar.

### DIFF
--- a/src/puppetlabs/ezbake/dependency_utils.clj
+++ b/src/puppetlabs/ezbake/dependency_utils.clj
@@ -130,7 +130,8 @@
         (if-let [project-jar-entry (find-file-in-jar project-jar
                                                      file-path)]
           (.getInputStream project-jar project-jar-entry)))
-      (log/warnf "Unable to find project jar file: '%s'" project-jar-file))))
+      (lein-main/abort
+       (format "Unable to find project jar file: '%s'" project-jar-file)))))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Public
--- a/src/puppetlabs/ezbake/dependency_utils.clj
+++ b/src/puppetlabs/ezbake/dependency_utils.clj
@@ -4,6 +4,7 @@
   (:require [cemerick.pomegranate.aether :as aether]
             [me.raynes.fs :as fs]
             [clojure.string :as str]
+            [clojure.tools.logging :as log]
             [leiningen.core.main :as lein-main]))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
@@ -116,6 +117,21 @@
         (assoc acc project-name (.getInputStream jar-file jar-entry)))
       acc)))
 
+(defn get-stream-from-project-jar
+  [lein-project file-path]
+  (let [project-jar-file (fs/file (:target-path lein-project)
+                                  (str
+                                   (:name lein-project)
+                                   "-"
+                                   (:version lein-project)
+                                   ".jar"))]
+    (if (fs/exists? project-jar-file)
+      (let [project-jar (JarFile. project-jar-file)]
+        (if-let [project-jar-entry (find-file-in-jar project-jar
+                                                     file-path)]
+          (.getInputStream project-jar project-jar-entry)))
+      (log/warnf "Unable to find project jar file: '%s'" project-jar-file))))
+
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Public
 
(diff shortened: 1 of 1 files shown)


camlow325:
Ping @shrug and @cprice404 for review.

camlow325:
I tested this with pe-puppetserver, which currently has an ezbake.conf in its resources directory, to see that the package which was built included the pre/post start actions I was expecting.  That all worked fine.

I also tested this by building pe-puppetserver with an older version of pe-puppet-server-extensions which has its own ezbake.conf file and saw that the content of the two files were aggregated - multiple pre/post start actions preserved - as I believe is expected.

src/puppetlabs/ezbake/dependency_utils.clj
```diff
@@ -116,6 +117,21 @@
         (assoc acc project-name (.getInputStream jar-file jar-entry)))
       acc)))
 
+(defn get-stream-from-project-jar
+  [lein-project file-path]
+  (let [project-jar-file (fs/file (:target-path lein-project)
```
camlow325: Not sure if there's a more idiomatic way to construct the path to the base project's jar file.  Open to alternate suggestions.

src/puppetlabs/ezbake/dependency_utils.clj
```diff
@@ -116,6 +117,21 @@
...
+                                   (:name lein-project)
+                                   "-"
+                                   (:version lein-project)
+                                   ".jar"))]
+    (if (fs/exists? project-jar-file)
+      (let [project-jar (JarFile. project-jar-file)]
+        (if-let [project-jar-entry (find-file-in-jar project-jar
+                                                     file-path)]
+          (.getInputStream project-jar project-jar-entry)))
+      (log/warnf "Unable to find project jar file: '%s'" project-jar-file))))
```
camlow325: Maybe we should bail here with an exception / kill the ezbake run instead?  I wasn't sure if there were any "normal" circumstances where this could happen.

camlow325:
@waynr Could you take a look at this?

If this looks sane, I'd like to get it merged and a new ezbake released with it.  As it currently stands, our pe-puppetserver packaging is broken for Ankeny right now and this should fix the problem we're having.  If this isn't going to work, though, we'll need to go with an alternative solution for now.

src/puppetlabs/ezbake/dependency_utils.clj
waynr: I can't think of any off the top of my head, but I also haven't dived deep into ezbake in almost a year.

src/puppetlabs/ezbake/dependency_utils.clj
waynr: I don't think there are such circumstances, ezbake should always run the `uberjar` task while staging:

waynr: whoops: https://github.com/puppetlabs/ezbake/blame/master/src/puppetlabs/ezbake/core.clj#L434

camlow325: Sorry, not very familiar with this code.  Does this mean that it will always run when staging?  Would it then always run when building?  Also, I think the way this is written right now, the code would be looking for the local project jar and not the uberjar.  Do you think that's the right way to go - or no?

waynr:
@camlow325 I think it would be useful to bail if the current project jar can't be since some projects may depend on being able to retrieve the conf file from their jar. If we don't bail then the build will proceed and it won't be until package installation or later that erroroneous behavior starts showing up. Failing early will ensure that we find bugs in this behavior before that happens. Otherwise this looks good to me.

src/puppetlabs/ezbake/dependency_utils.clj
waynr: Yes, the `build` sub task calls the `stage` subtask which in turn calls the method that is run for leiningen's `uberjar` task. I ran `lein ezbake stage` locally and saw that both the uberjar and the local project jar are created--I believe the `uberjar` task depends on the `jar` task.

camlow325:
@waynr Sounds good.  I added a commit, 8896468, to change the warning when a jar is not found to throwing an exception instead.  You good with merging this, then?

src/puppetlabs/ezbake/dependency_utils.clj
```diff
@@ -116,6 +117,23 @@
...
+                                   (:name lein-project)
+                                   "-"
+                                   (:version lein-project)
+                                   ".jar"))]
+    (if (fs/exists? project-jar-file)
+      (let [project-jar (JarFile. project-jar-file)]
+        (if-let [project-jar-entry (find-file-in-jar project-jar
+                                                     file-path)]
+          (.getInputStream project-jar project-jar-entry)))
+      (throw
```
waynr: For leiningen it is better to use `leiningen.core.main/abort` as shown here: https://github.com/puppetlabs/ezbake/blob/master/src/puppetlabs/ezbake/exec.clj#L35

camlow325: Thanks, that makes sense.  Fixed that up in 757c2dc.

waynr:
:+1:

PR #289, opened 2015-11-03
(PE-12468) Increase service start timeout to 300 seconds

shrug:
Currently, in our testing environment we are occasionally
seeing services timeout on startup at boot. This increases
the default startup timeout to 300 seconds (from 180).

There are a few ways to try and resolve this, this is the most expedient to
try to resolve intermittent issues in testing.


stahnma:
This makes me sad. 

:+1:

underscorgan:
This also makes me sad.

PR #293, opened 2016-02-02
(RE-6278) Update permissions for "projconfdir".

geoffnichols:
Prior to this commit, Ezbake packages install the
'projconfdir' with 'root:root' ownership and
'755' permissions.

We should probably remove read/execute permissions
to this directory for the 'other/world' group, but
preserve the same permissions for project service
accounts.

This commit updates Ezbake to (1) install the
'projconfdir' with '750' permissions and (2) set
user/group ownership to a project's specific
user and group.

### DIFF
--- a/resources/puppetlabs/lein-ezbake/template/foss/ext/redhat/ezbake.spec.erb
+++ b/resources/puppetlabs/lein-ezbake/template/foss/ext/redhat/ezbake.spec.erb
@@ -17,6 +17,7 @@
 # https://github.com/puppetlabs/puppet-specifications/blob/af82509/file_paths.md
 %global _app_prefix /opt/puppetlabs/server/apps/<%= EZBake::Config[:real_name] %>
 %global _app_data /opt/puppetlabs/server/data/<%= EZBake::Config[:real_name] %>
+%global _projconfdir /etc/puppetlabs/<%= EZBake::Config[:real_name] %>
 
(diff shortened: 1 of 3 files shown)


shrug:
As long as none of the services need to read anything in each other's config dirs, this is a reasonable fix.

shrug:
ghostbust 265h

stahnma:
Could somebody do an installation of PE and ensure that we're not using group permissions to read other files? I don't think we are, but I'd rather not break everything and then find out.

geoffnichols:
I'm testing this now...

geoffnichols:
Tested this as follows:
* Used beaker to install latest pre-release build of PE 2015.3.x (2015.3.2-rc0-90-g1b4dcc7).
* Simulated the change by manually setting permission to 750 and changing ownership as follows: 
```
 chmod 750 /etc/puppetlabs/{puppetserver,orchestration-services,console-services,puppetdb}
 chown pe-puppet:pe-puppet /etc/puppetlabs/puppetserver
 chown pe-orchestration-services:pe-orchestration-services /etc/puppetlabs/orchestration-services
 chown pe-console-services:pe-console-services /etc/puppetlabs/console-services
 chown pe-puppetdb:pe-puppetdb /etc/puppetlabs/puppetdb
```
* Rebooted master.
* After the reboot, confirmed I could successfully log into to the PE Console and execute "Run Puppet"

stahnma:
:+1:

PR #295, opened 2016-02-15
(EZ-59) Pass mirrors setting to aether

cprice404:
This commit improves the places where ezbake calls into
pomegranate/aether to make sure that we are passing through
the value for the `mirrors` setting from the project (or
from the user profile), so that aether will look for artifacts
in our nexus repository before going out to clojars or
maven central.  This should improve performance and reduce
network I/O for our CI jobs, and help prevent inflating
our clojars download counts from CI jobs.

This change seems like an improvement and step forward regardless
of the version of leiningen that is being used, but until
technomancy/leiningen#2089 is resolved, there may still be
portions of our CI jobs that incorrectly ignore `:local-repo` /
`:mirrors` settings from the user profile.

### DIFF
--- a/src/puppetlabs/ezbake/dependency_utils.clj
+++ b/src/puppetlabs/ezbake/dependency_utils.clj
@@ -4,7 +4,6 @@
   (:require [cemerick.pomegranate.aether :as aether]
             [me.raynes.fs :as fs]
             [clojure.string :as str]
-            [clojure.tools.logging :as log]
             [leiningen.core.main :as lein-main]))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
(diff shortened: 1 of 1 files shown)


cprice404:
@waynr does this look OK to you?

waynr:
@cprice404 yeah, this looks good to me, :+1:

cprice404:
@shrug I'm not really sure who to ask for a second review on this these days :)

shrug:
:+1:

cprice404:
@shrug thanks!  Any idea when the next ezbake release might be?  I think that this change might resolve some transient test failures that various teams have been experiencing.

shrug:
We can cut one immediately if we need to. There are a couple of other ezbake tickets that I'd like to get in the next sprint or so, but if this is causing immediate pain we can do a release now and then another once those other tickets get in.

cprice404:
@shrug nope, if you were planning a release in the next week or so, I'm happy to wait.

PR #296, opened 2016-02-18
(EZ-61) add restart=on-failure

McdonaldSeanp:
To remedy some reoccuring issues with services needing a restart
this comit will add restart=on-failure and startlimitburst=5
to the service file for rpms. refer to further discussion at
https://tickets.puppetlabs.com/browse/RE-5962

Note that on-failure has been available in systemd since v12 [https://github.com/systemd/systemd/releases/tag/v12, systemd/systemd@50caaed]
and is supported on all the systemd platforms we support

### DIFF
--- a/resources/puppetlabs/lein-ezbake/template/foss/ext/redhat/ezbake.service.erb
+++ b/resources/puppetlabs/lein-ezbake/template/foss/ext/redhat/ezbake.service.erb
@@ -8,6 +8,8 @@ EnvironmentFile=/etc/sysconfig/<%= EZBake::Config[:project] %>
 User=<%= EZBake::Config[:user] %>
 TimeoutStartSec=<%= EZBake::Config[:start_timeout] %>
 TimeoutStopSec=60
+Restart=on-failure
+StartLimitBurst=5
 
 <% unless EZBake::Config[:redhat][:pre_start_action].empty? -%>
 PermissionsStartOnly=true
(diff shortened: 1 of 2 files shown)


haus:
(on-failure has been available in systemd since v12 [https://github.com/systemd/systemd/releases/tag/v12, https://github.com/systemd/systemd/commit/50caaedb2c416c8972f9b2f835a69a20c524e81e])

McdonaldSeanp:
@haus was that just a comment for documentation purposes or was that a question?

shrug:
I was curious about that as well. Without putting words in Haus' mouth, it's important to know that the systemd directives are available in all of the platforms we're supporting.

haus:
@McdonaldSeanp exactly what @shrug said. i wanted to make sure sles12, el7 and fedora all supported the directive.

McdonaldSeanp:
yep, and all three have versions of systemd wildly farther along than 12, so we should be gtg =)

melissa:
My one thought is that ticket should probably have been an ezbake ticket since this project is public, and it would be nice to have the ticket available for the public to reference as well

mckern:
:+1: pending a corresponding EZ Bake ticket. I'd use that ticket number instead of the RE ticket number for the commits and the PR.

McdonaldSeanp:
Done, ready for merge

PR #301, opened 2016-04-20
(EZ-76) Add sleep to wait_for_app

underscorgan:
Ports sometimes are open before the app is really ready to accept
connections. Adding a 5s sleep after we detect the port is open to avoid
race conditions.


underscorgan:
Built puppetserver with this change, was able to test that it slept for 5 seconds after the port was open before returning from `service puppetserver start`

McdonaldSeanp:
seems reasonable 👍

PR #305, opened 2016-04-27
(EZ-72) Support split bootstrap in configuration

jpinsonault:
This commit adds support to ezbake for splitting a project's bootstrap files
between `/etc/puppetlabs/{project}/services.d/` and
`/opt/puppetlabs/server/apps/{project}/config/services.d/`

This serves the ultimate goal of putting user-configurable service entries
in one location, and the services that should not be modified in another.

It accomplishes this by adding two new settings to ezbake which can be
specified in the `:lein-ezbake` section of a `project.clj`
* `:split-bootstraps` - A boolean to cause ezbake to tell trapperkeeper that
  it should look for it's bootstrap files in the two directories mentioned
  above.
* `:private-config-dir` - A directory in the project, who's contents will be
  layed down in the /opt directory mentioned above.

This change is backward compatible. If `:split-bootstraps` is false (the default), ezbake
will continue to work the same as before, expecting a "bootstrap.cfg" in
`/etc/puppetlabs/{project}/bootstrap.cfg`.

**The names of the settings are definitely not the best, please suggest better ones!**


resources/puppetlabs/lein-ezbake/template/global/install.sh.erb
```diff
@@ -117,8 +118,19 @@ function task_install {
     install -m 0755 ext/ezbake-functions.sh "${DESTDIR}${app_prefix}"
     install -m 0644 ext/ezbake.manifest "${DESTDIR}${app_prefix}"
     install -d -m 0755 "${DESTDIR}${projconfdir}/conf.d"
+<% if EZBake::Config[:split_bootstraps] -%>
```
haus: seems worth putting this install call into the conditional below to keep them together.

cprice404: agree

src/puppetlabs/ezbake/core.clj
```diff
@@ -99,14 +102,12 @@
 (defn relativize
   "Convert an absolute File to a relative File"
   [base-path absolute-file]
-  (let [base-file (if (instance? File base-path)
-                    base-path
-                    (File. base-path))]
```
cprice404: just curious what happened here?

cprice404:
LGTM other than the thing @haus mentioned and my one minor question; also naming stuff, so I'll go reply to that email thread now :)

resources/puppetlabs/lein-ezbake/template/global/install.sh.erb
```diff
@@ -27,6 +27,7 @@ symbindir=${symbindir:=/opt/puppetlabs/server/bin}
 app_prefix=${app_prefix:=/opt/puppetlabs/server/apps/${real_name}}
 app_data=${app_data:=/opt/puppetlabs/server/data/${real_name}}
 app_logdir=${app_logdir:=/var/log/puppetlabs/${real_name}}
+private_config_dir=${private_config_dir:=${app_prefix}/config}
```
cprice404: @shrug @melissa @stahnma is this where you'd expect this directory to end up?  e.g. `/opt/puppetlabs/server/apps/puppetserver/config/services.d/`?  I know that some apps (razor maybe?) have like a 'share' directory underneath that path.  Easy to change it right now if you have any preferences.

src/puppetlabs/ezbake/core.clj
jpinsonault: Ah, I was just trying to figure out how this function worked and in the process figured I'd clean it up a bit. Probably should have put it in a separate maint commit

jpinsonault:
Addressed feedback/renamed everything. I went with `:file` and `:directory` for the enum values, but I can change that super easy if we'd prefer `:bootstrap-cfg` and `services-d` or whatever else

resources/puppetlabs/lein-ezbake/template/global/install.sh.erb
cprice404: @stahnma any opinions on this path?

src/puppetlabs/ezbake/core.clj
```diff
@@ -406,7 +426,8 @@ Dependency tree:
                                                           :main-namespace
                                                           "puppetlabs.trapperkeeper.main")
          :java-args                 (get-local-ezbake-var lein-project :java-args
-                                                          "-Xmx192m")}))))
+                                                          "-Xmx192m")
+         :bootstrap-source          (get-local-ezbake-var lein-project :bootstrap-source :file)}))))
```
cprice404: Reading through the code, it kind of seems like if a user put in any value besides 'directory' for this setting, it'd just silently ignore it and default to `file`.  Seems like maybe we should wrap a little schema validation around this line and add a test that covers that.

Also, I don't know what others think, but I think I'd kind of prefer `bootstrap-cfg` and `services-d` for the legal values here.  They're not entirely intuitive, but I don't know that these values are either.  `directory` seems kind of misleading since if you pick it, you're actually getting set up with 2 directories.  I don't have a super strong opinion about this, though, so I'm ok with it like this, but would like to see if @nathanielksmith or anyone else has thoughts.

jpinsonault: Agreed on validation.

Yeah I was really on the fence, but I think you're right, directory just doesn't make sense

jpinsonault: One thing that's come up is that having the hyphens in the settings means we have to do some conversion when they get passed to the ruby scripts

haus: That's true for most of the other config settings already 😢 . java-args, terminus-files...

jpinsonault:
I've gone ahead and changed the names. Also added some validation about the allowed values. It's a little gross (especially the test), but I'm not sure how to disentangle the mess of arguments and side effects that these functions all have. Suggestions welcome

src/puppetlabs/ezbake/core.clj
nathanielksmith: :+1: on `bootstrap-cfg` and `services-d`, the parsing in rubyside seems worth it

.lein-failures
```diff
@@ -0,0 +1 @@
+#{}
```
cprice404: this prolly ain't sposeda be here, right?

cprice404: maybe add this file to the project's .gitignore?

src/puppetlabs/ezbake/core.clj
```diff
@@ -427,7 +442,8 @@ Dependency tree:
                                                           "puppetlabs.trapperkeeper.main")
          :java-args                 (get-local-ezbake-var lein-project :java-args
                                                           "-Xmx192m")
-         :bootstrap-source          (get-local-ezbake-var lein-project :bootstrap-source :file)}))))
+                                    ; Convert to string so ruby doesn't barf on the hyphens
+         :bootstrap-source          (name (get-local-ezbake-var lein-project :bootstrap-source :bootstrap-cfg))}))))
```
cprice404: any reason not to call your validation function here instead of during the action?  seems like a good practice to keep config initialization and validation as close together as possible?

resources/puppetlabs/lein-ezbake/template/pe/ext/default.erb
```diff
@@ -13,7 +13,14 @@ USER="<%= EZBake::Config[:user] %>"
 GROUP="<%= EZBake::Config[:group] %>"
 INSTALL_DIR="/opt/puppetlabs/server/apps/<%= EZBake::Config[:real_name] %>"
 CONFIG="/etc/puppetlabs/<%= EZBake::Config[:real_name] %>/conf.d"
+
+# Bootstrap path
+<% if EZBake::Config[:bootstrap_source] == 'services-d' -%>
+BOOTSTRAP_CONFIG="/etc/puppetlabs/<%= EZBake::Config[:real_name] %>/services.d/,${INSTALL_DIR}/config/services.d/"
```
stahnma: I'm sure this is legacy, but what's the point of config being the path name? `etc` kind of implies that.

stahnma:
👍

resources/puppetlabs/lein-ezbake/template/global/install.sh.erb
```diff
@@ -27,6 +27,7 @@ symbindir=${symbindir:=/opt/puppetlabs/server/bin}
 app_prefix=${app_prefix:=/opt/puppetlabs/server/apps/${real_name}}
 app_data=${app_data:=/opt/puppetlabs/server/data/${real_name}}
 app_logdir=${app_logdir:=/var/log/puppetlabs/${real_name}}
+system_config_dir=${system_config_dir:=${app_prefix}/config}
```
cprice404: @stahnma so you are good with, e.g., `/opt/puppetlabs/server/apps/puppetserver/config` for the "hidden" config files?  Just wanted to make absolutely sure before we merged it.

resources/puppetlabs/lein-ezbake/template/pe/ext/default.erb
jpinsonault: There's two paths in there with a comma, the second one is under `/opt/puppetlabs/server/apps/{project}/config`. I included `config/`just to make it clear, since like you've pointed out, `/opt` isn't where you usually see these kinds of files. I think there's a desire to potentially move some of the integral `conf.d` files that users shouldn't be editing under /opt as well

cprice404:
Test seems OK to me.  Just a few minor comments on the latest commit, then I'll be +1.

src/puppetlabs/ezbake/core.clj
jpinsonault: I did that initially, but testing it would require calling this `generate-ezbake-config-file`, and creating all the parameters that need to be passed in. I guess it wouldn't be much worse. The other reason is that putting it in `action` causes the stage command to fail immediately, before doing all the file copying. Saves a good 10 seconds, but yeah, probably better to move it back here

cprice404: Hmm... yeah, that sucks.  But, I expect it to be a rare enough thing, and easy enough to fix, that it seems like code maintainability probably trumps those other concerns.  As long as it doesn't make the test *too* hard to write.

cprice404: And if the test is super sucky then I'd probably be ok with just doing a unit-level test of the validation function?

jpinsonault:
@cprice404 I untangled a small amount of the templating code to make it testable. I almost feel like the test should just be for the validation function, but either way I think the code's better now than before

cprice404:
+1, anyone else want to take a final look at this before we merge it?

PR #306, opened 2016-05-04
(EZ-64) Add Xenial as a default build target

melissa:


geoffnichols:
It's been a while since I ezbaked. Do we also need to add a Xenial build target for PE?

melissa:
Yes, we likely should. However, I'm uncomfortable doing so until we have the majority of ezbake projects building and testing for xenial.

Luckily, we don't need new and different enterprise cows for that work. The same cows are used for both FOSS and PE

geoffnichols:
👍

PR #307, opened 2016-05-05
(EZ-48) Add support for debian systemd

melissa:


resources/puppetlabs/lein-ezbake/template/foss/ext/debian/ezbake.service.erb
```diff
@@ -0,0 +1,41 @@
+[Unit]
+Description=<%= EZBake::Config[:project] %> Service
+After=syslog.target network.target
+
+[Service]
+Type=simple
+EnvironmentFile=/etc/sysconfig/<%= EZBake::Config[:project] %>
```
haus: Is this right for debian? Usually debian has it in /etc/default.

resources/puppetlabs/lein-ezbake/template/foss/ext/debian/postrm.erb
```diff
@@ -1,5 +1,9 @@
 #!/bin/sh -e
 
+if [ -x "/lib/systemd/system/<%= EZBake::Config[:project] -%>.service ] ; then
```
haus: missing closing " from the filename.

resources/puppetlabs/lein-ezbake/template/foss/ext/debian/prerm.erb
```diff
@@ -1,9 +1,16 @@
...
+  systemctl --no-reload disable <%= EZBake::Config[:project] -%>.service > /dev/null 2>&1 || :
+  systemctl stop <%= EZBake::Config[:project] -%>.service > /dev/null 2>&1 || :
 else
+  # Using sysv
+  if [ "$1" = "failed-upgrade" ]; then
+    invoke-rc.d <%= EZBake::Config[:project] -%> stop || :
+  else
     if [ -x "/etc/init.d/<%= EZBake::Config[:project] -%>" ] || [ -e "/etc/init/<%= EZBake::Config[:project] -%>.conf" ]; then
-        invoke-rc.d <%= EZBake::Config[:project] -%> stop || exit $?
+      invoke-rc.d <%= EZBake::Config[:project] -%> stop || exit $?
```
haus: Interesting that in the sysv case we have `|| exit $?` and for systemd it's just `||:`

resources/puppetlabs/lein-ezbake/template/global/install.sh.erb
```diff
@@ -70,9 +71,15 @@ function task_service {
         else
             task install_source_rpm_systemd
         fi
-    elif [ "$OSFAMILY" = "Debian" ]; then
+      elif [ "$OSFAMILY" = "Debian" ]; then
+        sysv_codenames=("squeeze" "wheezy" "lucid" "precise" "trusty")
+        codename=`facter os.distro.codename`
```
haus: We aren't guaranteed to have facter at build time. Also, I believe this task never gets called by the packaging, but is just around for ezbake source installs.

haus: The rpm spec directly calls the correct sysv/systemd install.sh function, and I believe previously the debian services were installed automatically by just evaluating the template to the right filename in the debian directory.

resources/puppetlabs/lein-ezbake/template/foss/ext/debian/prerm.erb
melissa: I have no idea why

resources/puppetlabs/lein-ezbake/template/global/install.sh.erb
melissa: facter is used to derive `$OSFAMILY`, so in this case, I assume it's safe to run

resources/puppetlabs/lein-ezbake/template/foss/ext/debian/ezbake.service.erb
melissa: I belive in this case, yes, it is correct. We're not installing anything to /etc/default for debian in ezbake

puppet-agent also has this same directory structure https://github.com/puppetlabs/puppet/blob/master/ext%2Fsystemd%2Fpuppet.service#L7

resources/puppetlabs/lein-ezbake/template/global/install.sh.erb
haus: It's only used that way for source installs though. That code path is not invoked for the native package builds afaik.

haus: facter is only used in the task_service and task_all functions, which deb and rpm packaging never calls, either directly or indirectly. the source install does use this though i think.

resources/puppetlabs/lein-ezbake/template/foss/ext/debian/ezbake.service.erb
melissa: nope, you were right that it should be /etc/default

resources/puppetlabs/lein-ezbake/template/foss/ext/debian/postinst.erb
```diff
@@ -4,12 +4,16 @@
...
 if [ $1 = 'configure' -a -n $2 ] ; then
-  if [ -x '/etc/init.d/<%= EZBake::Config[:project] %>' ] ; then
-    # Using sysv
-    invoke-rc.d <%= EZBake::Config[:project] %> try-restart || :
-  elif [ -x /bin/systemctl ] ; then
+  if [ -e "/lib/systemd/system/<%= EZBake::Config[:project] -%>.service" ] ; then
     # Using systemd
+    if [ -e '/etc/init.d/<%= EZBake::Config[:project] %>' ] ; then
+      # We need to clean up the old sysv files if they still exist
+      rm '/etc/init.d/<%= EZBake::Config[:project] %>'
```
haus: So this might delete the init script on each package upgrade right? Like you delete it here, but will the next package upgrade reinstall the file?

resources/puppetlabs/lein-ezbake/staging-templates/project_data.yaml.mustache
```diff
@@ -25,6 +25,8 @@ templates:
     target: ext/debian/{{{project}}}.default
   - source: ext/debian/ezbake.init.erb
     target: ext/debian/{{{project}}}.init
+  - source: ext/debian/ezbake.service.erb
+    target: ext/debian/{{{project}}}.service
```
haus: Is putting the .service file in the debian directory enough for it to be installed automatically?

resources/puppetlabs/lein-ezbake/template/foss/ext/debian/postinst.erb
```diff
@@ -4,12 +4,16 @@
...
+    fi
     systemctl daemon-reload >/dev/null 2>&1 || :
-    systemctl try-restart <%= EZBake::Config[:project] %>.service
+    systemctl try-restart <%= EZBake::Config[:project] %>.service ||:
```
melissa: This doesn't seem to be restarting the service effectively

resources/puppetlabs/lein-ezbake/staging-templates/project_data.yaml.mustache
melissa: no, it's being install in the install shell script

resources/puppetlabs/lein-ezbake/template/foss/ext/debian/postinst.erb
melissa: The next package upgrade shouldn't reinstall the file if it's a systemd platform

resources/puppetlabs/lein-ezbake/template/foss/ext/debian/postinst.erb
melissa: So, I'm not getting consistent behavior on upgrade :( the service state isn't being preserved

resources/puppetlabs/lein-ezbake/template/foss/ext/debian/postrm.erb
melissa: fixed

resources/puppetlabs/lein-ezbake/template/foss/ext/debian/prerm.erb
melissa: I blame @shrug

resources/puppetlabs/lein-ezbake/staging-templates/project_data.yaml.mustache
haus: it's being installed by dh_installinit as part of the build, not install.sh (`dh_installinit -ppuppetserver` is called in the build).

https://wiki.debian.org/Teams/pkg-systemd/Packaging#systemd_unit_files_naming_and_installation

resources/puppetlabs/lein-ezbake/template/foss/ext/debian/postinst.erb
haus: They are both in the package:

```
dpkg --contents /opt/jenkins-builds/puppetserver/2.4.0.master.SNAPSHOT.2016.05.05T2334/artifacts/deb/xenial/PC1/puppetserver_2.4.0.mT.2016.05.05T2334puppetlabs1_all.deb | grep 'init.d'
drwxr-xr-x root/root         0 2016-05-05 16:39 ./etc/init.d/
-rwxr-xr-x root/root      4662 2016-05-05 16:35 ./etc/init.d/puppetserver
matthaus@saturn  /home/matthaus> dpkg --contents /opt/jenkins-builds/puppetserver/2.4.0.master.SNAPSHOT.2016.05.05T2334/artifacts/deb/xenial/PC1/puppetserver_2.4.0.master-0.1SNAPSHOT.2016.05.05T2334puppetlabs1_all.deb | grep '\.service'
-rw-r--r-- root/root       866 2016-05-05 16:35 ./lib/systemd/system/puppetserver.service
```

haus: So with both in the package, this will also have the downside of nuking the sysv script from all deb packages in the postinstall.

resources/puppetlabs/lein-ezbake/staging-templates/project_data.yaml.mustache
melissa: Gah! I noted that we were calling install.sh in the rules file, so I assumed it was doing the right thing :( this sucks

resources/puppetlabs/lein-ezbake/template/foss/ext/debian/postinst.erb
melissa: which we definitely don't want...

resources/puppetlabs/lein-ezbake/template/foss/ext/debian/postinst.erb
melissa: It looks like puppet-agent is always a runtime dependency of ezbake packages. I'm going to follow @shrug 's advice and record the state of the service we are upgrading using puppet, and then apply that specific state to ensure consistency

haus: It's only a dependency of the terminus packages, not the main package itself.

haus:
I think it might make more sense to borrow pe-java's approach https://github.com/puppetlabs/enterprise-dist/blob/2016.2.x/packaging/debian/java/debian/rules#L35 and in the rules file use the codenames to conditionally call install_sysv or install_sysd. Then only one of the two will end up in the final package and the post/pre scripts should work better. This also has the advantage of better mirroring the rpm spec approach.
This would mean renaming the files in the debian directory so that dh_installinit didn't find them or overriding the dh_installinit step in the rules.

resources/puppetlabs/lein-ezbake/template/foss/ext/debian/postinst.erb
melissa: shiit

underscorgan:
With updates at https://github.com/underscorgan/ezbake/commit/dc72d9d4c054104cd81a31944d2f28796a96cc46

I was able to get packages at http://builds.puppetlabs.lan/puppetserver/2.4.0.master.SNAPSHOT.2016.05.06T1415

I haven't had a chance to test those packages at all though (other than making sure systemd/sysvinit scripts get packaged correctly)

melissa:
```
# dpkg --contents precise/puppetserver_2.4.0.master-0.1SNAPSHOT.2016.05.07T1540puppetlabs1_all.deb | grep "init.d"
drwxr-xr-x root/root         0 2016-05-07 08:47 ./etc/init.d/
-rwxr-xr-x root/root      4662 2016-05-07 08:47 ./etc/init.d/puppetserver
# dpkg --contents jessie/puppetserver_2.4.0.master-0.1SNAPSHOT.2016.05.07T1540puppetlabs1_all.deb | grep "init.d"
```

Successful update to packaging, now just to figure out preserving the service state

resources/puppetlabs/lein-ezbake/template/foss/ext/debian/postinst.erb
melissa: Fixed!

resources/puppetlabs/lein-ezbake/template/foss/ext/debian/postinst.erb
melissa: I have no idea the best way to fix this :(

melissa:
Hmm, so, any thoughts on preserving service state on upgrade without taking advantage of puppet?

haus:
@melissa what happens on upgrade if the service was running with sysv and it becomes systemd and a restart happens?

melissa:
@haus a system restart, or `systemctl daemon-reload`? On system restart, it's still inactive, but loaded. No state change

melissa:
@haus @shrug thoughts?

resources/puppetlabs/lein-ezbake/template/foss/ext/debian/postinst.erb
```diff
@@ -4,12 +4,26 @@
...
+
+    # sometimes daemon-reload doesn't always do the right thing, especially if we're upgrading from a sysv service
```
melissa: nope, this isn't doing the right thing still

melissa:
Dag nabbit. So, it looks like I don't actually have access to anything here. Trying to preserver the state of the service prior to upgrade is proving impossible. We'd need to capture that information in the prerm of the old package (which I can't alter, since it's already out there). The soonest we can do that is in the preinst of the new package. By that point, the service has already been stopped if is had previously been running. This happens because we effectively always stop the service in the prerm https://github.com/puppetlabs/ezbake/blob/master/resources%2Fpuppetlabs%2Flein-ezbake%2Ftemplate%2Ffoss%2Fext%2Fdebian%2Fprerm.erb#L7


I'm at a loss as to what to do

haus:
@melissa I would recommend limiting the scope of this change to *just* xenial then. It won't be a break for anyone except users on nightlies, whom we can inform in the release announcements.

haus:
Then new platforms for deb and ubuntu can come in with systemd support and old platforms can eventually die off that don't have it.

melissa:
It will break things for people upgrading on Debian 8 as well.

But I've removed my attempts to preserve the state of the service on upgrade.

haus:
@melissa but if you didn't try to move debian 8 to systemd in this change it wouldn't break people. that's what i was getting at. Like in the rules add jessie to the sysv codenames.

melissa:
Ooh, gross! Okay, I've updated the PR. I've also changes the prerm script, so this should be easier to fix in the future

haus:
yea. totally gross and sadmaking, but also better for current jessie users for some definitions of better.

resources/puppetlabs/lein-ezbake/template/global/install.sh.erb
```diff
@@ -65,15 +68,18 @@ function task_service {
...
-        defaultsdir="/etc/sysconfig"
         if [ $MAJREV -lt 7 ]; then
             task install_source_rpm_sysv
         else
             task install_source_rpm_systemd
         fi
     elif [ "$OSFAMILY" = "Debian" ]; then
-        defaultsdir="/etc/default"
-        task install_source_deb
+        sysv_codenames=("squeeze" "wheezy" "lucid" "precise" "trusty")
```
haus: I think you'll want to add jessie here as well so source and package remain consistent.

haus:
I had one more comment, but otherwise this looks good to me.

resources/puppetlabs/lein-ezbake/template/global/install.sh.erb
melissa: addressed

melissa:
Updated!

haus:
:+1:

PR #308, opened 2016-05-06
(EZ-78) Docs for bootstrap work

jpinsonault:
This adds docs describing how to use the new ezbake settings

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,11 @@ configuration necessary.
   ; configuration files to copy into the staging directory.
   :config-dir "config"
 
+  ; If specified, this is a directory where lein-ezbake will look for additional
+  ; configuration files will end up under the install directory under /opt. It's
+  ; intended for configuration files that an end user should not edit
+  :system-config-dir "system-config"
+
   ; These variables are available to either modify the behavior of lein-ezbake
   ; in various ways or to populate values in template files.
   :vars {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -153,10 +158,17 @@ repository.
...
+By default, in the final packages produced by an ezbake build, there will be a "config" directory
 (usually `/etc/puppetlabs/<ezbake-project-name>`) that contains all of the final
 configuration files for the ezbake application.
 
+If the `:bootstrap-source` setting is set to `:services-d`, there will also be an additional
+configuration directory under the installation directory:
+`/opt/puppetlabs/server/apps/{project-name}/config/`
+This directory will then be used to hold a `services.d/` directory that contains bootstrap files.
+Usually this will be for bootstrap entries the application could not run without, and which should
+not be modified by the user. See [Bootstrap Files](#bootstrap_files)
```
cprice404: I think I might add a subsection markdown header before this paragraph, called something like "bootstrap-source".  Then explicitly mention the two legal values for the enum here.  It's fine to keep this paragraph brief and defer the more detailed explanation to the later section that you added, but I think you should at least call out the legal values for the setting.

README.md
```diff
@@ -206,6 +218,36 @@ The rule of thumb is:
...
   `config/conf.d/web-routes.conf`.
 
+### Bootstrap Files
+There are two ways of specifying a set of services to bootstrap your TK application with:
+
+1. A single `bootstrap.cfg` file in your project's `:config-dir` directory
+2. Two `services.d` directories which can contain any number of `*.cfg` files that will be
+   merged together
+
+The default is to use a single `bootstrap.cfg` directly under your `:config-dir`. Setting
```
cprice404: Would also mention the default value of the `:bootstrap-source` setting here, in case anyone prefers to be explicit in their config.

cprice404:
+1 other than my comments above.

jpinsonault:
@cprice404 Addressed your comments, how's that look?

PR #312, opened 2016-05-12
(maint) Fix defaultsdir and unitdir

underscorgan:
Variable substitution issues were causing both deb and el to get set to
the el values for defaultsdir and unitdir. This fixes that issue.


geoffnichols:
👍

PR #313, opened 2016-05-13
(maint) add the Ubuntu 16.10 (Xenial) cow to pe build defaults

McdonaldSeanp:
This will be required to build for 16.10 in ezbake projects


geoffnichols:
👍

PR #314, opened 2016-05-13
(maint) update name field in init scripts

McdonaldSeanp:
For debian init scripts we cannot use real_name because it resolves to the
wrong thing. change this behavior to use EZBake::Config[:project]


shrug:
👍

PR #316, opened 2016-05-19
(SERVER-1247) fix bug with services-d bootstrap on systemd

cprice404:
In systemd, it doesn't appear that variable interpolation is supported
inside of the defaults files.  Prior to this commit, when using
the `services-d` option for bootstrap configuration, we were generating
a defaults file that relied on interpolation.  This commit changes
the code to avoid the interpolation.

### DIFF
--- a/resources/puppetlabs/lein-ezbake/template/foss/ext/default.erb
+++ b/resources/puppetlabs/lein-ezbake/template/foss/ext/default.erb
@@ -16,7 +16,7 @@ CONFIG="/etc/puppetlabs/<%= EZBake::Config[:real_name] %>/conf.d"
 
 # Bootstrap path
 <% if EZBake::Config[:bootstrap_source] == 'services-d' -%>
-BOOTSTRAP_CONFIG="/etc/puppetlabs/<%= EZBake::Config[:real_name] %>/services.d/,${INSTALL_DIR}/config/services.d/"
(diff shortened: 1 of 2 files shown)


cprice404:
This hasn't been tested yet.  Hoping that Joe will have time to do that tomorrow.

haus:
Copypasta all the things

resources/puppetlabs/lein-ezbake/template/foss/ext/default.erb
```diff
@@ -11,12 +11,11 @@ JAVA_ARGS="<%= EZBake::Config[:java_args] %>"
 # These normally shouldn't need to be edited if using OS packages
 USER="<%= EZBake::Config[:user] %>"
 GROUP="<%= EZBake::Config[:group] %>"
-INSTALL_DIR="/opt/puppetlabs/server/apps/<%= EZBake::Config[:real_name] %>"
```
underscorgan: grepping for INSTALL_DIR it looks like this variable is used in other places and only set here, so may want to leave this defined and just not use interpolation in the BOOTSTRAP_CONFIG below?

cprice404: ah, ok, thanks.  Yeah that might explain the failure I'm seeing locally.  I'll try changing that and see if it fixes my local tests... thanks

cprice404:
squashed up a commit that hopefully addresses @underscorgan 's feedback; testing locally now.

cprice404:
OK, tested this locally, seems to be working now.  @underscorgan you ok with us merging this after I get one more review?

@haus xoxo

underscorgan:
👍

PR #320, opened 2016-06-28
(EZ-83) Update init scripts to set a UMASK of 027

sarameisburger:
Set the default privileges to rw-r----- (permissions 640) so files are not world readable, even if moved out of the properly permissioned parent directories.


shrug:
👍

PR #321, opened 2016-07-26
(EZ-86) Add configuration of open file limit to services

MikaelSmith:
Allows ezbake projects to configure a default open file limit as part of
service startup.

### DIFF
--- a/resources/puppetlabs/lein-ezbake/template/foss/ext/debian/ezbake.service.erb
+++ b/resources/puppetlabs/lein-ezbake/template/foss/ext/debian/ezbake.service.erb
@@ -12,6 +12,9 @@ Restart=on-failure
 StartLimitBurst=5
 #set default privileges to -rw-r-----
 UMask=027
+<% if EZBake::Config[:open_file_limit] -%>
+LimitNOFILE=<%= EZBake::Config[:open_file_limit] %>
+<% end -%>
 
 <% unless EZBake::Config[:debian][:pre_start_action].empty? -%>
 PermissionsStartOnly=true
--- a/resources/puppetlabs/lein-ezbake/template/foss/ext/redhat/init.erb
+++ b/resources/puppetlabs/lein-ezbake/template/foss/ext/redhat/init.erb
@@ -66,6 +66,10 @@ start() {
     # Move any heap dumps aside
     echo -n $"Starting $prog: "
 
+    <% if EZBake::Config[:open_file_limit] -%>
+    [ -n "$OPEN_FILE_LIMIT" ] && ulimit -n $OPEN_FILE_LIMIT
+    <% end -%>
+
     <% EZBake::Config[:redhat][:pre_start_action].each do |action| -%>
     <%= action %>
     <% end -%>
--- a/resources/puppetlabs/lein-ezbake/staging-templates/ezbake.rb.mustache
+++ b/resources/puppetlabs/lein-ezbake/staging-templates/ezbake.rb.mustache
@@ -5,6 +5,7 @@ module EZBake
       :user           => '{{{user}}}',
       :group          => '{{{group}}}',
       :start_timeout  => '{{{start-timeout}}}',
+      :open_file_limit => '{{{open-file-limit}}}',
       :uberjar_name   => '{{{uberjar-name}}}',
       :config_files   => [{{{config-files}}}],
       :system_config_files   => [{{{system-config-files}}}],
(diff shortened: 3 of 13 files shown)


MikaelSmith:
I think this would work to configure open file limit on all the platforms we produce service packaging for. I'm working on testing them.

MikaelSmith:
Ok, fixed up a few things. Tested on RedHat with systemd.

MikaelSmith:
I'm not sure if modifying the init scripts is redundant, when I'm also setting default/sysconfig. However others like START_TIMEOUT seem to be set both places.

geoffnichols:
I'd probably lean towards actually setting the prescribed open file limit in the init scripts/systemd service files. 

If we go with that approach, should we optionally allow a custom number of files parameter to be set as an environment variable in default/sysconfig?

MikaelSmith:
That probably makes more sense. So leave the ulimit calls in all the init/systemd scripts, but make it look for an env variable that's set in default/sysconfig?

geoffnichols:
Yes, that sounds right to me.

MikaelSmith:
Pushed up, working on testing it.

resources/puppetlabs/lein-ezbake/template/foss/ext/debian/ezbake.service.erb
```diff
@@ -12,6 +12,7 @@ Restart=on-failure
 StartLimitBurst=5
 #set default privileges to -rw-r-----
 UMask=027
+LimitNOFILE=${OPEN_FILE_LIMIT}
```
MikaelSmith: The environment variable doesn't appear to be available here :(

geoffnichols: Ah, ok. That seems to be consistent with [this note](https://lists.freedesktop.org/archives/systemd-devel/2015-September/034092.html).

MikaelSmith:
Squashed as I think it's ready. I'll try this out on RHEL 6/7 and Ubuntu 12.04/16.04.

resources/puppetlabs/lein-ezbake/template/foss/ext/redhat/init.erb
```diff
@@ -66,6 +66,8 @@ start() {
     # Move any heap dumps aside
     echo -n $"Starting $prog: "
 
+    [ -z "$OPEN_FILE_LIMIT" ] && ulimit -n $OPEN_FILE_LIMIT
```
geoffnichols: I think we may also want erb conditional tag wrapping for these.

MikaelSmith: So we don't want to allow someone to configure these themselves from the defaults, unless we've specifically set a default?

geoffnichols: On second thought, it's probably fine without erb tag wrapping.

geoffnichols:
👍

geoffnichols:
👍

MikaelSmith:
They all passed.

resources/puppetlabs/lein-ezbake/template/foss/ext/debian/ezbake.service.erb
```diff
@@ -12,6 +12,9 @@ Restart=on-failure
 StartLimitBurst=5
 #set default privileges to -rw-r-----
 UMask=027
+<% if EZBake::Config[:open_file_limit] -%>
```
haus: So because the value is wrapped in '' in ezbake.rb, it will *always* be non-nil/truthy, even if not set intentionally. This breaks anything not specifying the open_file_limit.

MikaelSmith: Oh, grr. So need to use `unless EZBake::Config[:open_file_limit].empty?`?

haus: @janelu2 has a fix that changes the default to "nil", which evals to nil in ruby, which, if unwrapped, should be falsey. That will involve touching less files and seems reasonable to me.

haus: fix is here is you want to see: https://github.com/puppetlabs/ezbake/pull/325

PR #322, opened 2016-07-28
(EZ-84) Don't create empty pidfile if there's no pid

geoffnichols:
Prior to this commit, the EL init script's "find_my_pid" function
creates an empty pidfile if the service is not running.

This commit updates the "find_my_pid" function to check that $pid is not
a zero-length string before writing a pidfile.


ScottGarman:
:+1:

PR #323, opened 2016-07-28
(EZ-68) add reload capability

janelu2:
Details in [EZ-68](https://tickets.puppetlabs.com/browse/EZ-68) 

This PR creates a `reload` CLI command that allows for a file containing a counter to be incremented each time a `reload` is called (HUP signal).


cprice404:
@underscorgan and/or @shrug : this is not scheduled to ship for Davis but we'd like to start working on getting it in.  This PR would provide cli commands for all of the ezbake apps, e.g.:

`puppetserver reload`

which would send a HUP and then wait for the service to be restarted.  Once we get this merged, we will basically want to re-work @haus other PR (#283) so that it calls this command rather than trying to do the HUP directly.

In addition to code review for this PR we have a couple of timing issues / challenges to sort out:

1. This depends on a commit in the master branch of TK that hasn't been released yet.  We can probably do that release whenever, as TK 1.5.0.  The new functionality there is backward-compatible and opt-in.
2. Once this PR goes into an ezbake release we *probably* don't want that release to be consumed by any clojure apps that aren't on TK 1.5.0, because otherwise they'll end up with a {{reload}} cli command that won't work properly.
3. I assume that we'll want this PR, plus another one that @janelu2 will be filing shortly that adds a similar `puppetserver start` command, plus the PR that hooks `reload` into the service frameworks, plus a PR that hooks `start` into the service frameworks, to all come out in the same ezbake release together.
4. I assume we probably don't want that ^^ ezbake release to block other ezbake changes that may be needed for Davis.

Given all of that I'm thinking maybe it would make sense for us to move some branches around in this repo.  Perhaps:

* create a new `0.2.x` branch off of the current `stable` branch
* merge the current `master` branch down to `stable` and use that for Davis
* change the version number on the new master branch to 0.5.x or 0.6.x or something, and that way we can safely start to allow this series of commits to land there?

WDYT?

resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+restartfile=/opt/puppetlabs/server/data/<%= EZBake::Config[:real_name] %>/restartcounter.txt
```
cprice404: @shrug @underscorgan I think we've agreed that this file path should be ok, but there's one thing that is a little wonky here right now, which is that we also need to put this path into the `global.conf` file in order for everything to work.  That file is managed in the individual clojure projects, not in ezbake, so if we do it this way we're sorting of relying on the devs to use this exact path in their packaging.  Another option might be for us to make this path a config parameter on ezbake itself, and then the devs would still have to specify the value in two places but both of those places would be in their own repos.  Any thoughts?

resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
```diff
@@ -0,0 +1,24 @@
...
+
+if [ -e $restartfile ];  then
+    kill -HUP $(pgrep -f <%= EZBake::Config[:uberjar_name] %>)
+    while [ "$cur" == "$initial" ] ;do
+        sleep 0.1
+        cur=$(head -1 $restartfile)
+
+        ((timeout--))
+        if [ "$timeout" = 0 ]; then
+            exit 1
```
haus: Probably worth adding a message here about timing out (and what the timeout was).

resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
```diff
@@ -0,0 +1,24 @@
...
+        fi
+    done
+else
+    echo "Restart file does not exist"
+    exit 1
```
cprice404: If we somehow ended up in a state where a project was consuming this version of ezbake but did *not* opt-in to the Trapperkeeper 'restart-file' feature, then the command (e.g. `puppetserver reload`) would always end up logging this error.  I *think* that probably seems like the right way to go, though we could change the error message to something more generic, like `Application 'puppetserver' does not support 'reload'.`, since we don't want users to ever need to know anything about these restart files.

The other option would be to have the script go ahead and send the HUP if the restart file doesn't exist, but then just exit immediately rather than waiting for the server to come up to ready, since it won't have a way of knowing about that.  I think I prefer the hard fail but am interested in hearing opinions from @underscorgan @shrug @haus or whomever else.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+restartfile=/opt/puppetlabs/server/data/<%= EZBake::Config[:real_name] %>/restartcounter.txt
+cur=$(head -1 $restartfile)
```
haus: I'm not sure the `head -1` behavior should be relied upon. `head -n 1 $restartfile` is more explicit and follows the flags that head advertises in the man page.

janelu2: this has been changed

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+restartfile=/opt/puppetlabs/server/data/<%= EZBake::Config[:real_name] %>/restartcounter.txt
+cur=$(head -1 $restartfile)
+initial=$cur
+timeout="300"
+source /etc/sysconfig/puppetserver
+
+if [ -e $restartfile ];  then
+    /usr/bin/java $JAVA_ARGS -Djava.security.egd=/dev/urandom -cp "${INSTALL_DIR}/puppet-server-release.jar" clojure.main -m puppetlabs.trapperkeeper.main --config "${CONFIG}" -b "${BOOTSTRAP_CONFIG}" &
```
cprice404: @underscorgan this is where we copied the line from your other script.  The rest of this file looks fairly different because our mechanism for detecting when the service has started has changed... if this is missing anything else that we need from your script please let us know.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
```diff
@@ -0,0 +1,25 @@
...
+    while [ "$cur" == "$initial" ] ;do
+        sleep 0.1
+        cur=$(head -1 $restartfile)
```
cprice404: @janelu2 you have a couple more `head -1` occurrences in this file.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
```diff
@@ -0,0 +1,25 @@
...
+
+        ((timeout--))
+        if [ "$timeout" = 0 ]; then
+            exit 1
+        fi
+    done
+else
+    echo "Restart file does not exist"
```
cprice404: Same question here for @shrug / @underscorgan / @haus  as what I asked in the other file.  I don't want to keep that TCP port loop around, so I kind of think we should just say that no one should ever use this `start` script unless they've turned on the restart file configuration, and so it's desirable to fail hard right here.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
janelu2: :hankey:

janelu2:
Just a heads up, the PR for the `reload` CLI is in [this PR](https://github.com/puppetlabs/ezbake/pull/326) @underscorgan @cprice404 @shrug or anyone else

resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
underscorgan: +1 for failure, and rewording it as @cprice404 suggested.

shrug:
@cprice404 That branching plan makes sense to me.

cprice404:
@shrug thanks.  Would it be OK for me to go ahead and take care of the branching stuff now-ish, then, so that we can keep this work moving?  Is there anything I'd need to wait for or be careful of, timing-wise, before taking action on that branching plan?

And do you have any opinions about what version number the new `master` branch should be?  If I make it `0.5.x` then we have to be careful about if anything comes up that would require an `0.5.0` for Davis?

Maybe I should just call the master branch `1.0.0` since we've been using this in production for quite a while now?

resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
shrug: @cprice404: @underscorgan and I were just talking about this. I'm not sure we fully understand the second option (the ezbake option).

shrug:
@cprice404 Yes, I think this can move forward. Assuming we don't see any other breaking changes in the short term (I don't have anything pending, the gem stuff should be purely additive), then 1.0.0 probably makes sense. Otherwise, yeah 0.4.x or something?

cprice404:
@shrug OK I'll send out an email to the dev list with a branching proposal, and if it doesn't get any negative feedback I'll do the branching stuff tomorrow-ish.

janelu2:
@shrug @cprice404 https://github.com/puppetlabs/puppetserver/pull/1160 This is the global.conf file that is modified to have a restart-file path

resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
cprice404: Sorry, that wasn't very clear on my part.  So here's an example of a change that we'd need to make in the Puppet Server repo in order to opt-in to the "restart file" pattern:

https://github.com/puppetlabs/puppetserver/pull/1160/files#diff-f5caea6fde69fc958f8bb5d6f8605f97R5

We have to get that path into our configuration.  But then we also need to get the path into these start/reload scripts.  In the current implementation, we are duplicating the `/opt/puppetlabs/server/data` prefix here in the EZBake repo and there, in the Puppet Server repo.  And we'd basically be expecting/requiring any other consumers of this functionality to do it the same way.

So, the other option that I thought of was to add a new EZBake config option called `restart-file`, which would go somewhere like here:

https://github.com/puppetlabs/puppetserver/blob/master/project.clj#L87-L95

In that case, the line above in this PR would just get the full path from `EZBake::Config`, and the `/opt/puppetlabs/server/data` part wouldn't need to appear in this PR.  We'd still have the path hard-coded in two places, but they'd both be in the Puppet Server repo and there wouldn't be anything in ezbake that forced people to choose a specific path.

The more I think about it, the more I think that that is probably a better way to go.  Let me know if you agree, and we can re-work this PR.

shrug: Gotcha, that makes sense. 👍  to the ezbake config option.

janelu2:
Also reworked `start.erb` to check if file does *not* exist and create it if it does not. It will run the start command afterwards.

janelu2:
@cprice404 @haus any comments on the rework of `start.erb`?

resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
```diff
@@ -0,0 +1,25 @@
...
+        cur=$(head -n 1 $restartfile)
+
+        ((timeout--))
+        if [ "$timeout" = 0 ]; then
+            echo "Reload timed out"
+            exit 1
+        fi
+    done
+else
+    echo "Application 'puppetserver' does not support 'reload'"
```
haus: probably don't want to hard code puppetserver here.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+restartfile=<%= EZBake::Config[:restart_file] %>
+timeout="300"
+
+if [ -e $restartfile ];  then
```
haus: you'll want to wrap $restartfile in double quotes here to ensure you don't get a bash syntax error if restartfile is not set.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+restartfile=<%= EZBake::Config[:restart_file] %>
+timeout="300"
+
+if [ ! -e $restartfile ];  then
```
haus: you'll want to wrap $restartfile in double quotes here

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
```diff
@@ -0,0 +1,26 @@
...
+    echo "1" > $restartfile
```
haus: what if you can't write to the file?

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+restartfile=<%= EZBake::Config[:restart_file] %>
+timeout="300"
```
haus: should timeout here use the start-timeout ezbake var instead of being hardcoded to 300?

haus:
@janelu2 you'll want to wrap all of the bash uses of $restartfile in double quotes. If restart-file has special characters in it, trying to write to it or read from it without double quoting it will cause lots of odd problems.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
cprice404: Should be able to use that 'real_name' template variable here.

cprice404:
LGTM other than @haus ' suggestions.

@shrug @underscorgan now that we have an isolated `master` branch for ezbake we'll probably go ahead and merge this once @janelu2 has addressed the feedback.  Let us know if you see anything else or have any other reasons why we should wait.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
janelu2: should all instances of $restartfile be wrapped in double quotes?

haus: @janelu2 yes (https://github.com/puppetlabs/ezbake/pull/323#issuecomment-237979632)

janelu2: @haus just saw that. facepalm sorry lol

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
```
haus: I'd recommend either adding `set -e` here or adding `-e` to the bash invocation. `-e` means any non-zero exit will halt the script, which can be good, for example if on line 13 we don't have permissions to create a new directory.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
```diff
@@ -0,0 +1,33 @@
...
+    echo "1" > "$restartfile"
+fi
+
+${JAVA_BIN} ${JAVA_ARGS} -Djava.security.egd=/dev/urandom -cp ${INSTALL_DIR}/<%= EZBake::Config[:uberjar_name] %> clojure.main -m <%= EZBake::Config[:main_namespace] %> --config "${CONFIG}" -b "${BOOTSTRAP_CONFIG}" &
+
+cur=$(head -n 1 "$restartfile")
+initial=$cur
+
+while [ "$cur" == "$initial" ] ;do
+    sleep 0.1
```
haus: With this sleep as .1 seconds, the timeout doesn't mean what it used to (300 second timeout would become 30 seconds). I'd recommend sleeping a full second here, as start is quite a bit slower than reload will be.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
```diff
@@ -0,0 +1,33 @@
...
+    cur=$(head -n 1 "$restartfile")
+
+    ((timeout--))
+    if [ "$timeout" = 0 ]; then
+        echo "Startup timed out"
```
haus: I'd recommend including the timeout here. eg `"Startup timed out after <%= EZbake::Config[:start_timeout] %> seconds"`.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+restartfile=<%= EZBake::Config[:restart_file] %>
+timeout=<%= EZBake::Config[:start_timeout] %>
```
haus: I think it's okay to use the start timeout here and just use 1/10 of it, as we are below in the while loop. Eventually it might make sense to break out into its own ezbake setting, but i have trouble imagining an app where reload time won't be proportional to startup time.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
```diff
@@ -14,7 +14,7 @@ if [ -r "$restartfile" ];  then
 
         ((timeout--))
         if [ "$timeout" = 0 ]; then
-            echo "Reload timed out after <%= EZbake::Config[:start_timeout] %> seconds"
+            echo "$timeout" | awk '{printf "Reload timed out after %.3f secs\n", $1/10}'
```
janelu2: @haus I was pondering changing the sleep to be a sleep variable so that I could pass it in to this instead of hardcoding the /10, but I wasn't sure if it was ever going to be changed again and thus not needed to be made into a reference. does this look ok as is to you?

haus: You'd need to be doing math either way. I mentioned this in another comment, but it may have gotten lost.

"I would recommend doing this in ruby over shell. <%= EZbake::Config[:start_timeout] / 10 %>"

janelu2: omg 

also i think that i did not see that comment when I made that change, github must not update very quickly for me

cprice404:
Remove the "FOR REVIEW" label - I'm +1 on this and haven't heard any objections from @shrug and @underscorgan , so, planning to merge it as soon as @haus is +1.

haus:
@cprice404 i think i'm good here.

PR #325, opened 2016-07-28
(MAINT) fix LimitNOFILE parse error

janelu2:
This PR fixes a parse error that was cause by `open_file_limit` being wrapped in quotes. This caused it to always be set which caused parse errors on systemd. This was fixed by removing the quotes in the mustache template and explicitly defining `nil` as a string rather than a nil value in core.clj. This allowed it to become nil once the mustache template is evaluated.

### DIFF
--- a/resources/puppetlabs/lein-ezbake/staging-templates/ezbake.rb.mustache
+++ b/resources/puppetlabs/lein-ezbake/staging-templates/ezbake.rb.mustache
@@ -5,7 +5,7 @@ module EZBake
       :user           => '{{{user}}}',
       :group          => '{{{group}}}',
       :start_timeout  => '{{{start-timeout}}}',
-      :open_file_limit => '{{{open-file-limit}}}',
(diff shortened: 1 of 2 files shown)


janelu2:
@haus

haus:
@janelu2 A brief description of what was busted in the commit message would be rad. Something like...

"Because open_file_limit was wrapped in '', it was always set, which caused parse errors on systemd when the setting was not supplied. Explicitly defaulting to the string nil in clojure allows it to become a ruby nil once the mustache template is evaluated."

janelu2:
@haus hahah ok :)

janelu2:
@haus ok I added a description building on what you gave me. sorry, i usually don't crap out like that :hankey: got a bit lazy

haus:
:+1: from me. @shrug @geoffnichols either of you want a look here?

shrug:
👍

PR #326, opened 2016-07-28
(DO NOT MERGE) (EZ-90) add `start` subcommand for ezbake apps

janelu2:
NOTE: see conversation about branching considerations on #323 .

-------

This adds a `start` subcommand that will launch the java process asynchronously as a daemon and check for an increment of the counter in the restart file. Once the counter has been incremented, we can then know that the service has successfully started.

### DIFF
--- a/resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
+++ b/resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
@@ -6,7 +6,7 @@ initial=$cur
 timeout="300"
 
 if [ -e $restartfile ];  then
-    /usr/bin/java $JAVA_ARGS -Djava.security.egd=/dev/urandom -cp ${INSTALL_DIR}/<%= EZBake::Config[:uberjar_name] %> clojure.main -m <%= EZBake::Config[:main_namespace] %> --config "${CONFIG}" -b "${BOOTSTRAP_CONFIG}" &
+    ${JAVA_BIN} ${JAVA_ARGS} -Djava.security.egd=/dev/urandom -cp ${INSTALL_DIR}/<%= EZBake::Config[:uberjar_name] %> clojure.main -m <%= EZBake::Config[:main_namespace] %> --config "${CONFIG}" -b "${BOOTSTRAP_CONFIG}" &
     while [ "$cur" == "$initial" ] ;do
         sleep 0.1
         cur=$(head -n 1 $restartfile)
--- a/resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
+++ b/resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+
+restartfile=/opt/puppetlabs/server/data/<%= EZBake::Config[:real_name] %>/restartcounter.txt
+cur=$(head -1 $restartfile)
+initial=$cur
+timeout="300"
+source /etc/sysconfig/puppetserver
+
+if [ -e $restartfile ];  then
+    /usr/bin/java $JAVA_ARGS -Djava.security.egd=/dev/urandom -cp "${INSTALL_DIR}/puppet-server-release.jar" clojure.main -m puppetlabs.trapperkeeper.main --config "${CONFIG}" -b "${BOOTSTRAP_CONFIG}" &
+    while [ "$cur" == "$initial" ] ;do
+        sleep 0.1
+        cur=$(head -1 $restartfile)
+
+        ((timeout--))
+        if [ "$timeout" = 0 ]; then
+            echo "Startup timed out"
+            exit 1
+        fi
+    done
+else
+    echo "Restart file does not exist"
+    exit 1
+fi
+
+exit 0
--- a/resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
+++ b/resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
@@ -1,16 +1,15 @@
 #!/usr/bin/env bash
 
 restartfile=/opt/puppetlabs/server/data/<%= EZBake::Config[:real_name] %>/restartcounter.txt
-cur=$(head -1 $restartfile)
+cur=$(head -n 1 $restartfile)
 initial=$cur
 timeout="300"
-source /etc/sysconfig/puppetserver
 
 if [ -e $restartfile ];  then
-    /usr/bin/java $JAVA_ARGS -Djava.security.egd=/dev/urandom -cp "${INSTALL_DIR}/puppet-server-release.jar" clojure.main -m puppetlabs.trapperkeeper.main --config "${CONFIG}" -b "${BOOTSTRAP_CONFIG}" &
+    /usr/bin/java $JAVA_ARGS -Djava.security.egd=/dev/urandom -cp ${INSTALL_DIR}/<%= EZBake::Config[:uberjar_name] %> clojure.main -m <%= EZBake::Config[:main_namespace] %> --config "${CONFIG}" -b "${BOOTSTRAP_CONFIG}" &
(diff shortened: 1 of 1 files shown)


resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+restartfile=/opt/puppetlabs/server/data/<%= EZBake::Config[:real_name] %>/restartcounter.txt
+cur=$(head -1 $restartfile)
+initial=$cur
+timeout="300"
+source /etc/sysconfig/puppetserver
+
+if [ -e $restartfile ];  then
+    /usr/bin/java $JAVA_ARGS -Djava.security.egd=/dev/urandom -cp "${INSTALL_DIR}/puppet-server-release.jar" clojure.main -m puppetlabs.trapperkeeper.main --config "${CONFIG}" -b "${BOOTSTRAP_CONFIG}" &
```
haus: I think you've left some hard coded puppet-server bits in this template.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+restartfile=/opt/puppetlabs/server/data/<%= EZBake::Config[:real_name] %>/restartcounter.txt
+cur=$(head -1 $restartfile)
+initial=$cur
+timeout="300"
+source /etc/sysconfig/puppetserver
```
haus: This will already be sourced by the time the start command is running. start gets executed by a cli wrapper that lives here: https://github.com/janelu2/ezbake/blob/c5dd17967249f96efa81ff05a6415fa39ffb078f/resources/puppetlabs/lein-ezbake/staging-templates/cli-app.erb#L4-L17

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
haus: Also, you should be able to copy more of this line from https://github.com/janelu2/ezbake/blob/c5dd17967249f96efa81ff05a6415fa39ffb078f/resources/puppetlabs/lein-ezbake/template/global/ext/cli/foreground.erb#L8-L12 (things like which java to use, uberjar path are already defined there).

janelu2: would `restartfile`, the `source` call, and the `puppet-server-release.jar` all have to be modified?

haus: I think it should be safe to remove the source call entirely (I left a comment to that effect).
restartfile should be safe, unless it moves to an ezbake setting.

janelu2: cool, then the only thing that needs to be changed (other than removing `source) is the .jar file?

haus: @janelu2 if you look at the line i linked earlier in this comment, you can probably copy almost all of it and not need to make fixes to this line. I think the only thing missing from the line i linked is the `&` at the end.

haus: except you won't want the logappender bits.

haus: This comment from the other PR here https://github.com/puppetlabs/ezbake/pull/323#discussion_r72710963 makes me think I should back away from making more comments until the reload PR has had some eyes on it.

janelu2: wow sorry, i didn't see your other comments until after i wrote that comment. makes total sense.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+restartfile=/opt/puppetlabs/server/data/<%= EZBake::Config[:real_name] %>/restartcounter.txt
+cur=$(head -n 1 $restartfile)
+initial=$cur
+timeout="300"
+
+if [ -e $restartfile ];  then
+    /usr/bin/java $JAVA_ARGS -Djava.security.egd=/dev/urandom -cp ${INSTALL_DIR}/<%= EZBake::Config[:uberjar_name] %> clojure.main -m <%= EZBake::Config[:main_namespace] %> --config "${CONFIG}" -b "${BOOTSTRAP_CONFIG}" &
```
haus: `/usr/bin/java` needs to be replaced by a variable of some sort. there should be one defined in the default file and used in other commands.

janelu2: can `-Djava.security.egd=/dev/urandom` be replaced as well? or is that just a flag

haus: it's a flag. i don't recall why it's being set.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
```diff
@@ -0,0 +1,25 @@
...
+        cur=$(head -n 1 $restartfile)
+
+        ((timeout--))
+        if [ "$timeout" = 0 ]; then
+            echo "Startup timed out"
+            exit 1
+        fi
+    done
+else
+    echo "Restart file does not exist"
```
cprice404: Same question here for @shrug / @underscorgan / @haus as what I asked in the other file. I don't want to keep that TCP port loop around, so I kind of think we should just say that no one should ever use this start script unless they've turned on the restart file configuration, and so it's desirable to fail hard right here.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+restartfile=/opt/puppetlabs/server/data/<%= EZBake::Config[:real_name] %>/restartcounter.txt
+cur=$(head -n 1 $restartfile)
+initial=$cur
+timeout="300"
+
+if [ -e $restartfile ];  then
+    ${JAVA_BIN} ${JAVA_ARGS} -Djava.security.egd=/dev/urandom -cp ${INSTALL_DIR}/<%= EZBake::Config[:uberjar_name] %> clojure.main -m <%= EZBake::Config[:main_namespace] %> --config "${CONFIG}" -b "${BOOTSTRAP_CONFIG}" &
```
cprice404: @underscorgan this is where we copied the line from your other script. The rest of this file looks fairly different because our mechanism for detecting when the service has started has changed... if this is missing anything else that we need from your script please let us know.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
underscorgan: That makes sense to me!

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
underscorgan: This looks right. I didn't test this outside of running from init scripts with ExecStart/daemon though, so it may need some additional work if it's also intended for stand-alone work (not sure if it will correctly not kill the java process).

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
shrug: Same, 👍

janelu2:
Closing this PR and moving this script to [this PR](https://github.com/puppetlabs/ezbake/pull/323) to keep things on the same page.

PR #329, opened 2016-08-03
(MAINT) merge up from stable to master, 2016-08-03

cprice404:


cprice404:
This merge-up from stable to master ended up just being a few predictable places in docs.  Going to merge, but ping @camlow325 or @shrug or @underscorgan in case you are interested in double-checking afterward.

PR #330, opened 2016-08-03
(MAINT) merge down from master 2016-08-03

cprice404:
I've diffed this locally and the new stable has no diffs from the current master.


cprice404:
There were no merge conflicts when merging down from master to stable, so, merging this.  ping @camlow325 @shrug @underscorgan just in case.

PR #333, opened 2016-08-09
(EZ-68) fix typo `EZbake` to `EZBake`

janelu2:
This is a fixup PR, changes two typos in EZBake.

### DIFF
--- a/README_BRANCHING.md
+++ b/README_BRANCHING.md
@@ -1,5 +1,5 @@
 This document attempts to capture the details of our branching strategy
-for EZbake.
+for EZBake.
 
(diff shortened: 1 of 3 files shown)


janelu2:
@haus looking ok?

haus:
@janelu2 did you check the reload script for the same typo?

janelu2:
@haus yep, that was the only typo. Also needed to cast the timeout "300" to a float. This fix will be up sooooooooooooooooooooooooooooooooooooon

janelu2:
@haus ok, should be good to go now. do you see anything else?

haus:
nope. seems good.

PR #334, opened 2016-08-10
(EZ-69) Add support for service reload by calling new "reload" subcommand 

janelu2:
Previously, service reload was not supported. After adding support for reload in Trapperkeeper, this PR updates the init scripts to call the new `puppetserver reload` CLI instead. This reload will restart the service without exiting the JVM process and return a 0 exit code depending on whether or not the restart was successful.

### DIFF
--- a/resources/puppetlabs/lein-ezbake/template/foss/ext/debian/ezbake.init.erb
+++ b/resources/puppetlabs/lein-ezbake/template/foss/ext/debian/ezbake.init.erb
@@ -127,7 +127,6 @@ get_status_q()
     get_status >/dev/null 2>&1
 }
 
-
 #
 # Function that restarts the daemon/service
 #
@@ -154,13 +153,8 @@ do_restart()
 # Function that sends a SIGHUP to the daemon/service
 #
 do_reload() {
-     #
-     # If the daemon can reload its configuration without
-     # restarting (for example, when it is sent a SIGHUP),
-     # then implement that here.
-     #
-     start-stop-daemon --stop --signal HUP --quiet --pidfile $PIDFILE --name $NAME
-     return 0
+    /opt/puppetlabs/server/apps/${REALNAME}/bin/${REALNAME} reload
+    echo $?
 }
 
 case "$1" in
--- a/resources/puppetlabs/lein-ezbake/template/foss/ext/debian/ezbake.init.erb
+++ b/resources/puppetlabs/lein-ezbake/template/foss/ext/debian/ezbake.init.erb
@@ -150,6 +150,19 @@ do_restart()
     esac
 }
 
+#
+# Function that sends a SIGHUP to the daemon/service
+#
+do_reload() {
+     #
+     # If the daemon can reload its configuration without
+     # restarting (for example, when it is sent a SIGHUP),
+     # then implement that here.
+     #
+     start-stop-daemon --stop --signal HUP --quiet --pidfile $PIDFILE --name $NAME
+     return 0
+}
+
 case "$1" in
     start)
         [ "$VERBOSE" != no ] && log_daemon_msg "Starting $DESC" "$NAME"
@@ -175,16 +188,17 @@ case "$1" in
         get_status_q || exit 0
         do_restart
         ;;
-    restart|force-reload)
-        #
-        # If the "reload" option is implemented then remove the
-        # 'force-reload' alias
-        #
+    restart)
         [ "$VERBOSE" != no ] && log_daemon_msg "Restarting $DESC" "$NAME"
         do_restart
         ;;
+    force-reload|reload)
+        [ "$VERBOSE" != no ] && log_daemon_msg "Reloading $DESC" "$NAME"
+        do_reload
+        log_end_msg $?
+        ;;
     *)
-        echo "Usage: $SCRIPTNAME {start|stop|status|condrestart|try-restart|restart|force-reload}" >&2
+        echo "Usage: $SCRIPTNAME {start|stop|status|condrestart|try-restart|restart|force-reload|reload}" >&2
         exit 3
         ;;
 esac
--- a/resources/puppetlabs/lein-ezbake/template/foss/ext/debian/ezbake.init.erb
+++ b/resources/puppetlabs/lein-ezbake/template/foss/ext/debian/ezbake.init.erb
@@ -203,4 +203,4 @@ case "$1" in
         ;;
 esac
 
-:
+exit 0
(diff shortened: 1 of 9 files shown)


janelu2:
@haus @underscorgan @cprice404 I've tested this on systemv ubuntu, systemd ubuntu, and cent7. Going to test cent6 now. Does this look correct so far?

janelu2:
update: works fine on cent6 too

resources/puppetlabs/lein-ezbake/template/foss/ext/debian/ezbake.init.erb
```diff
@@ -150,6 +149,14 @@ do_restart()
     esac
 }
 
+#
+# Function that sends a SIGHUP to the daemon/service
+#
+do_reload() {
+    /opt/puppetlabs/server/apps/${REALNAME}/bin/${REALNAME} reload
+    echo $?
```
haus: you probably want a `return $?` instead of `echo $?` here.

resources/puppetlabs/lein-ezbake/template/foss/ext/debian/ezbake.init.erb
```diff
@@ -175,18 +182,19 @@ case "$1" in
...
-        # 'force-reload' alias
-        #
+    restart)
         [ "$VERBOSE" != no ] && log_daemon_msg "Restarting $DESC" "$NAME"
         do_restart
         ;;
+    force-reload|reload)
+        [ "$VERBOSE" != no ] && log_daemon_msg "Reloading $DESC" "$NAME"
+        do_reload
+        log_end_msg $?
```
haus: does `log_end_msg` exit the script? if not it would be worth capturing the return value of `do_reload` so that the script can exit with it after `log_end_msg`

resources/puppetlabs/lein-ezbake/template/pe/ext/debian/ezbake.init.erb
```diff
@@ -150,6 +150,14 @@ do_restart()
     esac
 }
 
+#
+# Function that sends a SIGHUP to the daemon/service
+#
+do_reload() {
+     /opt/puppetlabs/server/apps/${REALNAME}/bin/${REALNAME} reload
+     echo $?
```
haus: `echo` => `return`

resources/puppetlabs/lein-ezbake/template/pe/ext/debian/ezbake.init.erb
```diff
@@ -175,18 +183,19 @@ case "$1" in
...
-        # If the "reload" option is implemented then remove the
-        # 'force-reload' alias
-        #
+    restart)
         [ "$VERBOSE" != no ] && log_daemon_msg "Restarting $DESC" "$NAME"
         do_restart
         ;;
+    force-reload|reload)
+        [ "$VERBOSE" != no ] && log_daemon_msg "Reloading $DESC" "$NAME"
+        do_reload
```
haus: same comment as in the foss version. we should probably be capturing the return value here to exit with it below.

resources/puppetlabs/lein-ezbake/template/foss/ext/redhat/init.erb
```diff
@@ -131,6 +131,13 @@ rh_status_q() {
     rh_status >/dev/null 2>&1
 }
 
+reload() {
+    echo -n $"Reloading $prog: "
```
cprice404: Not that I have any issue with it, but I'm curious why there is an echo here when there is not one in most of the other scenarios.

resources/puppetlabs/lein-ezbake/template/foss/ext/debian/ezbake.init.erb
haus: alternatively, if `log_end_msg` returns whatever `$?` was passed in, then the exit at the end can just be `exit $?`

resources/puppetlabs/lein-ezbake/template/pe/ext/redhat/init.suse.erb
```diff
@@ -166,6 +166,12 @@ sl_status() {
     rc_status -v
 }
 
+reload() {
+    echo -n $"Reloading ${prog}:"
+    killproc -p "${PIDFILE}" "$prog" -HUP
```
haus: This needs updating to use the reload script.

resources/puppetlabs/lein-ezbake/template/pe/ext/redhat/init.suse.erb
```diff
@@ -166,6 +166,12 @@ sl_status() {
     rc_status -v
 }
 
+reload() {
+    echo -n $"Reloading ${prog}:"
```
cprice404: OK, so there are echos for this in some other places.  Maybe it's consistent.

resources/puppetlabs/lein-ezbake/template/foss/ext/redhat/init.erb
haus: https://github.com/puppetlabs/ezbake/pull/334/files/13da957138a585aa929bf79a7e9dcdf07158199c#diff-636dbe07137aa76779d2f7c2ab149076R67 and https://github.com/puppetlabs/ezbake/pull/334/files/13da957138a585aa929bf79a7e9dcdf07158199c#diff-636dbe07137aa76779d2f7c2ab149076R104 echo messages when starting and stopping, so this is consistent within the file.

cprice404:
This LGTM but I'm totally unqualified to be the one signing off on it :)

resources/puppetlabs/lein-ezbake/template/foss/ext/debian/ezbake.init.erb
janelu2: @haus I couldn't find any documentation on it so I wrote a silly script calling `log_end_msg` 

` do_stuff() {
    return 0
}

do_more() {
    return 3
}

do_stuff
log_end_msg $?`

do_more
log_end_msg $?

and the output is:
`...done.`
`...fail!`

It doesn't seem like it exits the script. Just logs whether or not the return is 0. It also does not return any value, so I'll capture the return value of `do_reload`

janelu2: actually, after reading more into this, I don't think that this switch statement is actually meant to return anything (unless case *). I believe the purpose of this case is simply to log whether or not the return from the `do_*` function calls are 0, so I think that what I have should be fine as is. Correct me if I'm wrong

haus:
I'm good here. 👍

cprice404:
I think we're ready to merge this, but would love to get a +1 from @underscorgan , @shrug , or @melissa .

mckern:
👍; this seems sane to me

PR #338, opened 2016-08-22
Maint/master/merge stable

melissa:


melissa:
I have no idea if this is the correct thing to do, but this commit merges recent changes from the stable branch into the master branch.

PR #339, opened 2016-08-26
(SERVER-366) Add boolean to disable logrotate

jpinsonault:
This commit adds a new boolean ezbake setting to disable logrotate in a project

SERVER-366 covers work to limit the amount of disk space puppetserver will use for its logs. We decided to accomplish that by utilizing the built in features of the logback library. Logback will
* Do daily rotating of logs
* Rotate the logs out once they reach a certain size
* Limit the total size of all the puppetserver log file

Logrotate is no longer needed, so we needed a way to disable logrotate in ezbake. The change in this PR is backwards compatible, as it defaults to turning logrotate on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,8 @@ configuration necessary.
     :group "puppet"
     :start-timeout "120"
     :build-type "foss"
-    :java-args "-Xms2g -Xmx2g -XX:MaxPermSize=256m"}}}
+    :java-args "-Xms2g -Xmx2g -XX:MaxPermSize=256m"
+    :logrotate-enabled true}}}
 
 ```
 
--- a/resources/puppetlabs/lein-ezbake/staging-templates/ezbake.rb.mustache
+++ b/resources/puppetlabs/lein-ezbake/staging-templates/ezbake.rb.mustache
@@ -43,5 +43,6 @@ module EZBake
                             "{{package}}" => "{{version}}",
                           {{/replaces-pkgs}}
                           },
+      :logrotate_enabled => {{{logrotate-enabled}}},
   }
 end
--- a/resources/puppetlabs/lein-ezbake/template/foss/ext/debian/rules.erb
+++ b/resources/puppetlabs/lein-ezbake/template/foss/ext/debian/rules.erb
@@ -20,6 +20,7 @@ export EZ_VERBOSE=1
 
 install/<%= EZBake::Config[:project] %>::
 	bash install.sh install_deb
+<% if EZBake::Config[:logrotate_enabled]-%>
 	if [[ ${DIST} == debian ]] ; then \
 		if [[ $$(echo "${RELEASE}>=7" | bc -l) -eq 1 || ${RELEASE} == testing || ${RELEASE} == unstable ]] ; then \
 			bash install.sh logrotate ; \
@@ -33,6 +34,7 @@ install/<%= EZBake::Config[:project] %>::
 			bash install.sh logrotate_legacy ; \
 		fi ; \
 	fi
+<% end -%>
 
 <% unless EZBake::Config[:terminus_info].empty? -%>
 install/<%= EZBake::Config[:project] %>-termini::
--- a/resources/puppetlabs/lein-ezbake/template/foss/ext/redhat/ezbake.spec.erb
+++ b/resources/puppetlabs/lein-ezbake/template/foss/ext/redhat/ezbake.spec.erb
@@ -98,11 +98,13 @@ env EZ_VERBOSE=1 DESTDIR=%{buildroot} prefix=%{_prefix} confdir=%{_sysconfdir} b
 env EZ_VERBOSE=1 DESTDIR=%{buildroot} prefix=%{_prefix} confdir=%{_sysconfdir} bindir=%{_bindir} rundir=%{_rundir}/<%= EZBake::Config[:project] %> defaultsdir=%{_sysconfdir}/sysconfig initdir=%{_initrddir} bash install.sh sysv_init_redhat
 %endif
 
+<% if EZBake::Config[:logrotate_enabled]-%>
 %if 0%{?fedora} >= 16 || 0%{?rhel} >= 7 || 0%{?sles_version} >= 12
 env EZ_VERBOSE=1 DESTDIR=%{buildroot} confdir=%{_sysconfdir} bash install.sh logrotate
 %else
 env EZ_VERBOSE=1 DESTDIR=%{buildroot} confdir=%{_sysconfdir} bash install.sh logrotate_legacy
 %endif
+<% end -%>
 
 <% unless EZBake::Config[:terminus_info].empty? -%>
 env EZ_VERBOSE=1 DESTDIR=%{buildroot} rubylibdir=%{rubylibdir} prefix=%{_prefix} bash install.sh termini
@@ -192,7 +194,9 @@ fi
 %endif
 %config(noreplace) %{_sysconfdir}/%{name}
 %config(noreplace) %{_sysconfdir}/sysconfig/%{name}
+<% if EZBake::Config[:logrotate_enabled]-%>
 %config(noreplace) %{_sysconfdir}/logrotate.d/%{name}
+<% end -%>
 <% if ! EZBake::Config[:cli_app_files].empty? -%>
 %{_bindir}/<%= EZBake::Config[:real_name] %>
 <% end -%>
--- a/resources/puppetlabs/lein-ezbake/template/pe/ext/debian/rules.erb
+++ b/resources/puppetlabs/lein-ezbake/template/pe/ext/debian/rules.erb
@@ -19,6 +19,7 @@ export EZ_VERBOSE=1
 
 install/<%= EZBake::Config[:project] %>::
 	bash install.sh install_deb
+<% if EZBake::Config[:logrotate_enabled]-%>
 	if [[ ${DIST} == debian ]] ; then \
 		if [[ $$(echo "${RELEASE}>=7" | bc -l) -eq 1 || ${RELEASE} == testing || ${RELEASE} == unstable ]] ; then \
 			bash install.sh logrotate ; \
@@ -32,6 +33,7 @@ install/<%= EZBake::Config[:project] %>::
 			bash install.sh logrotate_legacy ; \
 		fi ; \
 	fi
+<% end -%>
 
 <% unless EZBake::Config[:terminus_info].empty?  -%>
 install/<%= EZBake::Config[:project] %>-termini::
--- a/resources/puppetlabs/lein-ezbake/template/pe/ext/redhat/ezbake.spec.erb
+++ b/resources/puppetlabs/lein-ezbake/template/pe/ext/redhat/ezbake.spec.erb
@@ -152,11 +152,13 @@ env EZ_VERBOSE=1 DESTDIR=%{buildroot} prefix=%{_prefix} confdir=%{_sysconfdir} b
 %endif
 %endif
 
+<% if EZBake::Config[:logrotate_enabled]-%>
 %if 0%{?fedora} >= 16 || 0%{?rhel} >= 7 || 0%{?sles_version} >= 12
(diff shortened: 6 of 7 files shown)


cprice404:
LGTM.  @haus it looks like you did the original logrotate stuff?  Any idea who in RE we should ping for an RE review on this?

haus:
@jpinsonault the linked ticket doesn't seem to have anything to do with logrotate?

melissa:
The code looks find at first glance, but could you add more detail explaining why you are making this change?

jpinsonault:
@haus oops, thank you. Fixed. 

@melissa Yeah definitely. Updated the description

jpinsonault:
Also I should mention that there will be another PR for the master branch with the same changes

melissa:
Awesome, but we should also update the commit message itself with the new description.

Does it blend?

jpinsonault:
Updated the commit message

I don't have my Blendtec blender on hand unfortunately

melissa:
Fair :) but it builds packages with all the expected updates?

jpinsonault:
Haha, yeah I've done manual testing with the setting toggled on and off and observed logrotate being enabled and disabled on centos 6. I can try out more OSs though if you'd like

melissa:
Yeah, could you test a deb variety for me? Let's make sure debhelper doesn't decide to be helpful.

haus:
@jpinsonault what is the expected upgrade behavior here?

jpinsonault:
@haus What I would expect is that logrotate would be disabled. Other than that, the only side effect I'd expect is, if puppetserver has a large log file from before the upgrade, it will be rotated out once puppetserver restarts. The format of the timestamp on logrotate managed logs and the logs managed by logback will probably be different. 

I don't believe logback will recognize the old logrotate managed logs if the naming convention differs, and won't clean them up

jpinsonault:
@melissa Just tested on wheezy with it enabled and disabled, got the expected behavior

jpinsonault:
So I tried installing the version with logrotate disabled over the version with it enabled, and it in fact does not disable logrotate on upgrade. Not being very well versed in linux packaging, is there an acceptable, packaging friendly way of ensuring the `logrotate.d/puppetserver` file doesn't exist?

haus:
@jpinsonault what platform?

jpinsonault:
I tried on wheezy

haus:
@jpinsonault okay, yea, so this is going to be a difference between redhat and debian. on el or sles, i would expect one of 2 things to happen to the logrotate file.

1. if it is unmodified, it should be deleted
2. if it has been modified, it should be saved as /etc/logrotate.d/puppetserver.rpmsave or something (not sure if logrotate will skip loading this file or not, might be worth investigating).

On debian, all packaged files under /etc are considered conffiles, which means they aren't removed during a standard remove, but only during a purge. The default configuration for apt is such that package upgrades only do a remove, and not a purge (this is configurable). You can see here, that apt knows the logrotate file is obsolete, but has left it there until a purge happens.

```
root@o1beoc1lrl6s7ht:~# dpkg-query --showformat='${Conffiles}\n' --show puppetserver
 /etc/default/puppetserver 00f4fc6ef7c68d76474267f3e35e0fde
 /etc/puppetserver/logback.xml 7779b0bb9deb90cd8c9507ee09a4bbfc
 /etc/puppetserver/bootstrap.cfg bf3f18707d26e6149ea286bc92723df4
 /etc/puppetserver/conf.d/webserver.conf 6443ba5977dd89cacb2ca26d7e6511ca
 /etc/puppetserver/conf.d/puppetserver.conf 558554cbd9f461cebf020c6cff98d17a
 /etc/puppetserver/conf.d/os-settings.conf b19a8a0894bd27984e4ec6aab947b28a
 /etc/puppetserver/conf.d/ca.conf 733c38f57969885b832333eb00ddff43
 /etc/puppetserver/conf.d/web-routes.conf 2d1b2464c90748718bbeb1a29fb17536
 /etc/puppetserver/conf.d/global.conf 4f3f5a745c44e38e269f4cbfaeb9abb5
 /etc/puppetserver/request-logging.xml 5eee1ab8d5ea6215baceb592af6d962a
 /etc/init.d/puppetserver c1c88b05f234e74afb3b69db09d72579
 /etc/logrotate.d/puppetserver defadea6828bd62abad63d9510cfd44a obsolete
```

So for debian, if we want the file gone on package upgrade, we would have to add some logic to the postinst to rm the file if it exists and logrotate is disabled. Depending on how redhat behaves with a .rpmsave file inside of /etc/logrotate.d, we *might* want to do the same thing there. However, with this kind of solution, users would not be able to opt back into logrotate on their own as we would nuke it every time on package upgrade, so I would be wary of trying this unless we intend for logrotate to no longer be an option for our packages.

jpinsonault:
@haus I just tested an upgrade on el6 and it didn't remove or backup the logrotate file

The idea of blindly removing/moving the logrotate file on upgrade does make me a little uncomfortable. I'll discuss with Chris

haus:
@jpinsonault that's surprising. is there a machine up from the el6 effort?

jpinsonault:
@haus `li5pm13cpda89bx` is the machine

There's a few puppetserver rpms in the home dir. 

`puppetserver-1.1.4-0.1SNAPSHOT.2016.08.26T1601.el6.noarch.rpm` has logrotate disabled
`puppetserver-1.1.4-0.1SNAPSHOT.2016.08.29T1059.el6.noarch.rpm` is the same version, only with logrotate enabled
`puppetserver-1.1.5-0.1SNAPSHOT.2016.08.29T1309.el6.noarch.rpm` has logrotate disabled. I just bumped the version number to simulate an upgrade

haus:
@jpinsonault 

```
[root@li5pm13cpda89bx ~]# ls -l /etc/logrotate.d/
total 20
-rw-r--r--. 1 root root 103 Nov 29  2013 dracut
-rw-r--r--. 1 root root 256 Apr 25 11:52 puppet
-rw-r--r--. 1 root root 145 Aug 29 11:00 puppetserver
-rw-r--r--. 1 root root 210 Aug 15  2013 syslog
-rw-r--r--. 1 root root 100 Dec 17  2013 yum
```

```
[root@li5pm13cpda89bx ~]# rpm -Uvh puppetserver-1.1.5-0.1SNAPSHOT.2016.08.29T1309.el6.noarch.rpm
Preparing...                ########################################### [100%]
   1:puppetserver           ########################################### [100%]
[root@li5pm13cpda89bx ~]# ls -l /etc/logrotate.d/
total 16
-rw-r--r--. 1 root root 103 Nov 29  2013 dracut
-rw-r--r--. 1 root root 256 Apr 25 11:52 puppet
-rw-r--r--. 1 root root 210 Aug 15  2013 syslog
-rw-r--r--. 1 root root 100 Dec 17  2013 yum
```

haus:
And with an edited logrotate file:

```
[root@li5pm13cpda89bx ~]# rpm --force -Uvh puppetserver-1.1.4-0.1SNAPSHOT.2016.08.29T1059.el6.noarch.rpm
Preparing...                ########################################### [100%]
   1:puppetserver           ########################################### [100%]
[root@li5pm13cpda89bx ~]# vi /etc/logrotate.d/puppetserver
[root@li5pm13cpda89bx ~]# ls -l /etc/logrotate.d/
total 20
-rw-r--r--. 1 root root 103 Nov 29  2013 dracut
-rw-r--r--. 1 root root 256 Apr 25 11:52 puppet
-rw-r--r--. 1 root root 154 Aug 29 14:16 puppetserver
-rw-r--r--. 1 root root 210 Aug 15  2013 syslog
-rw-r--r--. 1 root root 100 Dec 17  2013 yum
[root@li5pm13cpda89bx ~]# rpm -Uvh puppetserver-1.1.5-0.1SNAPSHOT.2016.08.29T1309.el6.noarch.rpm
Preparing...                ########################################### [100%]
   1:puppetserver           ########################################### [100%]
warning: /etc/logrotate.d/puppetserver saved as /etc/logrotate.d/puppetserver.rpmsave
[root@li5pm13cpda89bx ~]# ls -l /etc/logrotate.d/
total 20
-rw-r--r--. 1 root root 103 Nov 29  2013 dracut
-rw-r--r--. 1 root root 256 Apr 25 11:52 puppet
-rw-r--r--. 1 root root 154 Aug 29 14:16 puppetserver.rpmsave
-rw-r--r--. 1 root root 210 Aug 15  2013 syslog
-rw-r--r--. 1 root root 100 Dec 17  2013 yum
```

cprice404:
@jpinsonault if the logrotate config is left in place, and the new logback config is added, what will happen?

haus:
Oh, also of note: logrotate is configured to ignore `.rpmsave` files:

```
[root@li5pm13cpda89bx ~]# logrotate -d -v /etc/logrotate.conf
reading config file /etc/logrotate.conf
including /etc/logrotate.d
Ignoring puppetserver.rpmsave, because of .rpmsave ending
```

jpinsonault:
@haus ah, yep, I was just doing `rpm -i ...`. That's good then

@cprice404 I would imagine logrotate will move `puppetserver.log`, and logback will attempt to move it around the same moment (logback at least will rotate at midnight). I would hope that turns out to be no worse than having an empty backup, or a backup with slightly different filename than you were expecting.

cprice404:
@jpinsonault if that's the behavior then I think we are probably ok with release notes that explain under what circumstances logrotate would be left on.  If it's not a big PITA to edit the cron for logrotate and make it happen more frequently so that you can actually observe what happens that would be awesome.

jpinsonault:
@cprice404 it shouldn't be that hard. Looks like you can trigger logrotate manually

jpinsonault:
@cprice404 I have logrotate rotating every half second and curls hitting the status endpoint as fast as possible, with logback's max file size set to 1KB. So far there are no problems that I can see. Half the files are logrotate backups and half are logback backups. I can't really look at all the files it's producing, but there are no errors in the output of `puppetserver foreground` or anything like that

cprice404:
@jpinsonault that sounds good to me... in that case I think it'd be good to make sure you and @haus and @melissa are on the same page in terms of 'best effort' for what should happen to the logrotate config on various platforms after an upgrade, but I'm (personally) fine with just documenting anything else we need to in the release notes.

jpinsonault:
@haus @melissa Since it sounds like on el, the file will be removed/backed up automatically, it kind of seems like we should make debian match that behavior. We definitely intend for the puppetserver log files to be managed by logback. Doing that and/or documenting the changes to logback/upgrade experience seems good enough to me. Thoughts?

haus:
Logrotate will ignore dpkg-old, dpkg-new, and dpkg-new extensions, so I would mv the file if it exists in the postinst to the ignored extension. users can copy it back to a new name if they want to keep it around. we can document that part.

jpinsonault:
Ok, well we discussed with Rob, and we decided that it's less of a sin to have some duplicate log files on debian upgrades than it is to remove the choice from the user and forcefully disable logrotate every time puppetserver is installed.

So we'll handle documenting this quirk on debian in the release notes/docs for puppetserver, and this PR is good as it is (at least with respect to upgrades)

melissa:
@jpinsonault just to be clear, you've opted to not add in a script that migrates the old logs to dpkg-old (or some other extension) for debian upgrades?

I would vote for to add the extension, as it sounds like logrotate is already configured to handle that scenario, and that would make it easy for users to migrate back to what they previously had. What does logrotate do when you upgrade on debian to a state that isn't supposed to use logrotate, but still have the logrotate files with unmodified filenames?

jpinsonault:
@melissa Yeah, we're advocating to not make any effort to remove the obsolete logrotate file on debian upgrades.

The concern we have about just adding the extension is the possibility of a user wanting to add the logrotate file back for some reason. If we did that in a postinst step it would just move that file blindly on every subsequent upgrade. 

The only downside if a user doesn't see our release notes and doesn't remove that file themselves is that they'll have a few extra harmless log files

melissa:
Awesome, thanks for clarifying! I'm on board with these changes 👍

jpinsonault:
Woo!

cprice404:
OK that seems like a +1(-ish) from @haus , me, @shrug , @melissa ... that seems like quorum... so....

PR #341, opened 2016-08-31
(EZ-95) Merge up logback changes from 0.2.x to master

jpinsonault:
This is a merge up of changes in SERVER-366 from `0.2.x` to `master`

SERVER-366 covers work to limit the amount of disk space puppetserver will use
for its logs. We decided to accomplish that by utilizing the built in features
of the logback library. Logback will:
* Do daily rotating of logs
* Rotate the logs out once they reach a certain size
* Limit the total size of all the puppetserver log file

Logrotate is no longer needed, so we needed a way to disable logrotate in ezbake.
The change in this commit is backwards compatible, as it defaults to turning
logrotate on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,8 @@ configuration necessary.
     :group "puppet"
     :start-timeout "120"
     :build-type "foss"
-    :java-args "-Xms2g -Xmx2g -XX:MaxPermSize=256m"}}}
+    :java-args "-Xms2g -Xmx2g -XX:MaxPermSize=256m"
+    :logrotate-enabled true}}}
 
 ```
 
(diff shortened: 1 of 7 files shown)


jpinsonault:
ping @cprice404 @melissa Might as well keep bothering the same people

The changes are identical to the 0.2.x PR that was merged yesterday

I've tested the new setting in puppetserver set to true/false/omitted on el and debian and observed that logrotate is enabled/disabled as expected

cprice404:
@jpinsonault these changes look fine to me, but a question on mechanics - is this the result of a git merge operation?  Because it would definitely be valuable to have the lineage preserved up through the branches.

jpinsonault:
I cherrypicked it from the other branch and resolved the conflicts

jpinsonault:
I'll go ahead and do a merge instead

jpinsonault:
Updated the PR, it's now a merge up of the 0.2.x commit

jpinsonault:
Also I retested it manually to make sure nothing got borked in the merge

cprice404:
+1 from me

cprice404:
@melissa OK with you if we do an 0.4.z release after this gets merged?

melissa:
👍 go for it

PR #346, opened 2016-09-02
(EZ-96) Exit with appropriate codes for start/stop in Debian init script

camlow325:
This commit fixes a bug introduced by EZ-70 where the start and stop
actions in Debian init scripts would errantly return a 1 (failure) exit
code when the start/stop succeeded rather than returning a 0.

### DIFF
--- a/resources/puppetlabs/lein-ezbake/template/foss/ext/debian/ezbake.init.erb
+++ b/resources/puppetlabs/lein-ezbake/template/foss/ext/debian/ezbake.init.erb
@@ -146,18 +146,22 @@ case "$1" in
     start)
         [ "$VERBOSE" != no ] && log_daemon_msg "Starting $DESC" "$NAME"
         do_start
-        case "$?" in
+        RETVAL="$?"
+        case "$RETVAL" in
             0|1) [ "$VERBOSE" != no ] && log_end_msg 0 ;;
             2) [ "$VERBOSE" != no ] && log_end_msg 1 ;;
         esac
+        exit "$RETVAL"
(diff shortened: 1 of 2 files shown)


camlow325:
@cprice404, @haus - would appreciate if you could take a look at this.

I tested this out by using it in a puppetserver package built for Ubuntu 14.04 with the puppetserver smoke test suite.  Without this change, the following commands issued by the smoke test suite would previously fail:

~~~~
puppet resource service puppetserver ensure=stopped
puppet resource service puppetserver ensure=running
~~~~

haus:
❤️

cprice404:
:see_no_evil: but wouldn't mind getting a review from @melissa or @underscorgan

melissa:
This looks reasonable to me. There aren't similar changes required for redhat packages?

camlow325:
@melissa Yeah, I didn't have the same problem with the el scripts.  In that case, looks like they were already allowing the original exit code from the action - start, stop, etc. - to percolate back through to the `exit` command.

The Debian ones were only problematic because the exit code was getting reset from 0 to 1 when the `[ "$VERBOSE" != "no" ` check returned 1 (when "$VERBOSE" was actually "no").  We don't have any of the "$VERBOSE" stuff in the el scripts.

https://github.com/puppetlabs/ezbake/blob/3828947bc1216154b2acf228bc98bc6b89793d92/resources/puppetlabs/lein-ezbake/template/foss/ext/redhat/init.erb#L134-L159

underscorgan:
👍

PR #350, opened 2016-09-12
(maint) Use new key for PE build defaults

underscorgan:


melissa:
I think @shrug and I decided stable PE should stay the old key?

underscorgan:
@melissa @shrug stable is the version that should be used for davis though, right?

shrug:
I may have made a mistake about that, I forgot about the new 0.2 branch, lets circle up and figure that out in the morning

PR #351, opened 2016-09-13
(maint) Update foss projects to use new key

melissa:


underscorgan:
there was a verbal 👍  from @shrug

PR #352, opened 2016-09-13
Add maintainers as per policy

stahnma:


stahnma:
Merged during RE PR review meeting.

PR #357, opened 2016-09-15
(EZ-97) Start / reload commands should error immediately after Java process dies

camlow325:
This PR primarily handles terminating the "wait for" loops immediately if the Java process spawned by the start or reload commands dies.  Previously, the commands would wait for a full configured "timeout" to expire before terminating.  In the process of hooking this up, I found a few related bugs around start functionality which are addressed in this PR, including:

* If the start script is terminated - e.g., by systemctl under systemd - while still running, the associated Java process is now terminated as well.  Previously, the Java process could continue running, leaving the resulting service in an inconsistent state (depending upon timing) from the service framework's perspective.

* Handle an exit code of "1" under Debian init as a failure instead of as a success.  Previously, the result of a start attempt when the service failed to start was treated as "success".

* Make the Debian init script's use of the VERBOSE flag for reload actions consistent with that used for other actions.

* Some cleanup around ensuring that the service pidfile is removed when the Java process fails to be started, helping make service frameworks more consistently report "failure" when the service isn't started properly.

### DIFF
--- a/resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
+++ b/resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
@@ -23,7 +23,7 @@ if [ -r "$restartfile" ];  then
         cur=$(head -n 1 "$restartfile")
 
         ((timeout--))
-        if [ "$timeout" = 0 ]; then
+        if [ $timeout -eq 0 ]; then
             echo "Reload timed out after <%= EZBake::Config[:start_timeout].to_i / 10 %> seconds"
             exit 1
         fi
--- a/resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
+++ b/resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
@@ -1,16 +1,24 @@
 #!/usr/bin/env bash
-set -e
+set +e
 
 restartfile=<%= EZBake::Config[:restart_file] %>
 timeout=<%= EZBake::Config[:start_timeout] %>
 
 if [ -r "$restartfile" ];  then
     pid=$(pgrep -f <%= EZBake::Config[:uberjar_name] %>)
-    kill -HUP $pid
+    kill -HUP $pid >/dev/null 2>&1
+    if [ $? -ne 0 ]; then
+        echo "Service not running so cannot be reloaded" 1>&2
+        exit 1
+    fi
     cur=$(head -n 1 "$restartfile")
     initial=$cur
     while [ "$cur" == "$initial" ] ;do
         kill -0 $pid >/dev/null 2>&1
+        if [ $? -ne 0 ]; then
+            echo "Process $pid exited before reload had completed" 1>&2
+            exit 1
+        fi
         sleep 0.1
         cur=$(head -n 1 "$restartfile")
 
--- a/resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
+++ b/resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
@@ -10,7 +10,7 @@ if [ -r "$restartfile" ];  then
     cur=$(head -n 1 "$restartfile")
     initial=$cur
     while [ "$cur" == "$initial" ] ;do
-        kill -0 $pid 2>/dev/null
+        kill -0 $pid >/dev/null 2>&1
         sleep 0.1
         cur=$(head -n 1 "$restartfile")
 
--- a/resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
+++ b/resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
@@ -5,10 +5,12 @@ restartfile=<%= EZBake::Config[:restart_file] %>
 timeout=<%= EZBake::Config[:start_timeout] %>
 
 if [ -r "$restartfile" ];  then
-    kill -HUP $(pgrep -f <%= EZBake::Config[:uberjar_name] %>)
+    pid=$(pgrep -f <%= EZBake::Config[:uberjar_name] %>)
+    kill -HUP $pid
     cur=$(head -n 1 "$restartfile")
     initial=$cur
     while [ "$cur" == "$initial" ] ;do
+        kill -0 $pid 2>/dev/null
         sleep 0.1
         cur=$(head -n 1 "$restartfile")
 
--- a/resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
+++ b/resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
@@ -7,6 +7,8 @@ dir=$(dirname "$restartfile")
 realname="<%= EZBake::Config[:real_name] %>"
 PIDFILE="/var/run/puppetlabs/${realname}/${realname}.pid"
 
+rm -f "$PIDFILE"
+
 if [ ! -e "$restartfile" ]; then
     mkdir -p "$dir"
     echo -n "0" > "$restartfile"
--- a/resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
+++ b/resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
@@ -1,11 +1,56 @@
 #!/usr/bin/env bash
+pid=$(pgrep -f <%= EZBake::Config[:uberjar_name] %>)
+
 set -e
 
 restartfile=<%= EZBake::Config[:restart_file] %>
-timeout=<%= EZBake::Config[:start_timeout] %>
+start_timeout=<%= EZBake::Config[:start_timeout] %>
+stop_timeout=60
 dir=$(dirname "$restartfile")
 realname="<%= EZBake::Config[:real_name] %>"
 PIDFILE="/var/run/puppetlabs/${realname}/${realname}.pid"
+
+write_pid_file() {
+    if [ ! -d "/var/run/puppetlabs/${realname}" ]; then
+        mkdir -p /var/run/puppetlabs/${realname}
+        chown -R $USER:$GROUP /var/run/puppetlabs/${realname}
+    fi
+    echo $1 > $PIDFILE
+}
+
+terminate_java_process() {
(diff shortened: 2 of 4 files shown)


resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
```diff
@@ -5,10 +5,12 @@ restartfile=<%= EZBake::Config[:restart_file] %>
 timeout=<%= EZBake::Config[:start_timeout] %>
 
 if [ -r "$restartfile" ];  then
-    kill -HUP $(pgrep -f <%= EZBake::Config[:uberjar_name] %>)
+    pid=$(pgrep -f <%= EZBake::Config[:uberjar_name] %>)
+    kill -HUP $pid
     cur=$(head -n 1 "$restartfile")
     initial=$cur
     while [ "$cur" == "$initial" ] ;do
+        kill -0 $pid >/dev/null 2>&1
```
camlow325: Not sure if `kill -0 $pid` is the best approach for ensuring that a pid is still active but found several references on the webz - e.g., http://stackoverflow.com/questions/11012527/what-does-kill-0-pid-in-a-shell-script-do and http://stackoverflow.com/questions/13260974/how-to-check-if-process-is-running-in-linux - indicating that it is a reasonably portable approach.  It seems to work fine on the different OSes I tested - EL6 / 7, Ubuntu 12.04 / 14.04 / 16.04, SLES 11 / 12.  Happy to use something different if others have better suggestions.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
```diff
@@ -1,21 +1,59 @@
...
         chown -R $USER:$GROUP /var/run/puppetlabs/${realname}
     fi
-    echo $pid > $PIDFILE
+    echo $1 > $PIDFILE
 }
 
+terminate_java_process() {
+    set +e
+
+    echo "Startup script was terminated before completion" 1>&2
```
camlow325: Might have been able to use `killproc` here to eliminate a number of lines of code but I'm not sure we can assume it will be available (and works the same way?) across OSes.  Seems like we use `start-stop-daemon` in Debian in some cases.  Seems best to avoid putting any OS-specific stuff into these scripts.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
```diff
@@ -26,19 +64,33 @@ fi
...
+        echo "Startup timed out after $start_timeout seconds" 1>&2
         exit 1
     fi
 done
+set -e
 
-find_my_pid
-exit 0
\ No newline at end of file
+write_pid_file $pid
```
camlow325: This probably seems redundant with the `write_pid_file` line above - close to where the background process is launched.  In doing some edge-case testing around concurrent restart attempts, I found that for some service frameworks that the pidfile could be removed if a separate script instance was terminated, leading it to not be present after this point even though the service had successfully started.  This just seemed like a good sanity check.  Should be harmless, essentially a no-op in most cases.

camlow325:
@cprice404, @haus, @underscorgan, @melissa - would appreciate if at least of some of y'all could take a look at this.  Would definitely like to have a +1 from at least one RE person before potentially merging this.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
haus: we have used kill -0 internally as well, just don't do it on solaris please.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
haus: yea, that is not cross platform.

camlow325:
FYI, I tested these changes with full packages on EL6 / 7, Ubuntu 12.04 / 14.04 / 16.04, and SLES 12.  I did some spot checking of the commands added on SLES 11 but hadn't done a fuller test of those from a built package.  I also have some related Beaker tests that I hope to add into the puppetserver smoke test suite which should give some coverage of these in the OSS and PE puppetserver pipelines.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
camlow325: Cool, gtk.  We don't support Solaris as a master platform, so should be okay here, right?

haus: yup

haus: so `kill -0 $pid` will return non-zero if the process died for some reason, which will then halt the reload script (because of `set -e`. Should we print some error out in that case, or do we assume the application log will have the relevant details?

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
```diff
@@ -1,21 +1,59 @@
...
+terminate_java_process() {
+    set +e
+
+    echo "Startup script was terminated before completion" 1>&2
+    kill -TERM $pid >/dev/null 2>&1
+    sleep 0.1
+
+    timeout=$stop_timeout
+    kill -0 $pid >/dev/null 2>&1
+    while [ $? == 0 ] && [ "$timeout" != 0 ]; do
```
haus: you've got a mix of `== 0` and `= 0` here. Would recommend `-eq 0` and `ne 0` for integer comparisons.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
camlow325: Yeah, that's all what I've seen / am expecting.  I would expect logs to have other data about the failure.  I went back/forth on this a bit, too.  I'm open to echoing a specific error message out here if others think it would be valuable.  Probably would have to do a `set +e` and `set -e` around it so the error can be caught and message written.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
camlow325: Hmm, yeah, you're right.  I can change all of the integer comparisons in the script to `-eq` and `-ne`, as appropriate.  I like that syntax a little better too.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
```diff
@@ -26,19 +64,33 @@ fi
 
 ${JAVA_BIN} ${JAVA_ARGS} -Djava.security.egd=/dev/urandom -cp ${INSTALL_DIR}/<%= EZBake::Config[:uberjar_name] %> clojure.main -m <%= EZBake::Config[:main_namespace] %> --config "${CONFIG}" -b "${BOOTSTRAP_CONFIG}" &
 
+pid=$!
```
haus: Probably worth a note here explaining what `$1!` is.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
haus: seems fine to me.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
```diff
@@ -1,21 +1,59 @@
 #!/usr/bin/env bash
+pid=$(pgrep -f <%= EZBake::Config[:uberjar_name] %>)
+
 set -e
```
haus: I think you've added enough error handling to start, that we should probably just lose all of the `set -e` `set +e` calls. I don't think there are any spots left in this script that could return non-zero really.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
haus: I don't have strong opinions here. Both seem reasonable.

haus:
This looks good to me. I left some style nits, but otherwise I'm :+1:

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
camlow325: Yeah, I think you're right.  I'll remove them.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
camlow325: I went ahead and added the trace message in f7b5fe8.  Also removed the `set -e` from the entire script since exit codes are generally handled anyway so it didn't seem necessary anymore.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
camlow325: Makes sense.  I added a comment in 6ebd2c4.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
camlow325: Did this in 69671df.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
camlow325: Did this in 74f7249.

camlow325:
@haus I think I hit all of your comments with my latest commits.  Let me know if you see anything else.  Thanks!

camlow325:
I found one small bug where a "Java process not killed" message might be logged even if the process actually was killed during an aborted startup.  Addd 9b02de5 to address that.

camlow325:
@underscorgan, @melissa - I think @haus is +1 on this now.  Any chance you could review this soon?  I have another PR that I want to put up which builds on this one so would like to get this one in soon, if possible.  Thanks!

haus:
Yea, I'm :+1:

melissa:
👍  this looks fine to me as well

PR #358, opened 2016-09-15
(EZ-88) Add per-project build dependencies

shrug:
This adds per-project build dependencies to ezbake, allowing projects to declare build deps in the same was as runtime deps for packaging.

This also adds the contents of the project data directory to packaging, instead of just packaging the top-level data directory. This allows package builds to install files there and have them picked up by packaging.

### DIFF
--- a/resources/puppetlabs/lein-ezbake/template/foss/ext/redhat/ezbake.spec.erb
+++ b/resources/puppetlabs/lein-ezbake/template/foss/ext/redhat/ezbake.spec.erb
@@ -104,6 +104,9 @@ Requires:         %{open_jdk}
 # net-tools is required for netstat usage in service unit file
 # See: https://tickets.puppetlabs.com/browse/SERVER-338
 Requires:         net-tools
+<% EZBake::Config[:redhat][:additional_build_dependencies].each do |dep| %>
+BuildRequires:    <%= dep %>
+<% end %>
 <% EZBake::Config[:redhat][:additional_dependencies].each do |dep| %>
 Requires:         <%= dep %>
 <% end %>
--- a/resources/puppetlabs/lein-ezbake/template/foss/ext/redhat/ezbake.spec.erb
+++ b/resources/puppetlabs/lein-ezbake/template/foss/ext/redhat/ezbake.spec.erb
@@ -262,7 +262,7 @@ fi
 %{_sym_bindir}/<%= bin_file.sub(%r{^ext/bin/}, "") %>
 %{_ux_bindir}/<%= bin_file.sub(%r{^ext/bin/}, "") %>
 <% end -%>
-%dir %attr(0770, <%= EZBake::Config[:user] %>, <%= EZBake::Config[:group] %>) %{_app_data}
+%attr(-, <%= EZBake::Config[:user] %>, <%= EZBake::Config[:group] %>) %{_app_data}
 %dir %attr(0755, <%= EZBake::Config[:user] %>, <%= EZBake::Config[:group] %>) %{_app_rundir}
 
 <% unless EZBake::Config[:terminus_info].empty? -%>
--- a/resources/puppetlabs/lein-ezbake/template/pe/ext/debian/control.erb
+++ b/resources/puppetlabs/lein-ezbake/template/pe/ext/debian/control.erb
@@ -2,7 +2,10 @@ Source: <%= EZBake::Config[:project] %>
 Section: utils
 Priority: optional
 Maintainer: Puppet Labs <info@puppetlabs.com>
-Build-Depends: debhelper (>= 7.0.0~), cdbs, bc, mawk, lsb-release, puppet-agent
+Build-Depends: debhelper (>= 7.0.0~), cdbs, bc, mawk, lsb-release, puppet-agent<%=
+    if !EZBake::Config[:debian][:additional_build_dependencies].empty?
+      ", " + EZBake::Config[:debian][:additional_build_dependencies].join(", ")
+    end %>
 Standards-Version: 3.9.1
 Homepage: http://puppetlabs.com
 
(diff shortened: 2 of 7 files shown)


resources/puppetlabs/lein-ezbake/template/foss/ext/redhat/ezbake.spec.erb
```diff
@@ -262,7 +265,7 @@ fi
 %{_sym_bindir}/<%= bin_file.sub(%r{^ext/bin/}, "") %>
 %{_ux_bindir}/<%= bin_file.sub(%r{^ext/bin/}, "") %>
 <% end -%>
-%dir %attr(0770, <%= EZBake::Config[:user] %>, <%= EZBake::Config[:group] %>) %{_app_data}
+%attr(0770, <%= EZBake::Config[:user] %>, <%= EZBake::Config[:group] %>) %{_app_data}
```
underscorgan: should this still be 0770 if it isn't a dir?

melissa: I think this change could use a bit more explanation

shrug:
Test build (with some additional puppetserver changes) at http://jenkins-release.delivery.puppetlabs.net/job/puppetserver-packaging-2016-09-15-15-50-41-2.6.0.master.SNAPSHOT.2016.09.15T1550/

The failures are related to the puppetserver change to install gems, still working through that

shrug:
Here's a build against puppetserver master:

http://jenkins-release.delivery.puppetlabs.net/job/puppetserver-packaging-2016-09-15-17-49-01-2.6.0.master.SNAPSHOT.2016.09.15T1748/

resources/puppetlabs/lein-ezbake/template/foss/ext/redhat/ezbake.spec.erb
shrug: Updated to set the dir perms correctly when we install, and not touch the files here

shrug: With this change there now may be files installed under that directory, so we need to pick them up

shrug:
This is intended to support a change like https://github.com/puppetlabs/puppetserver/compare/master...shrug:maint/master/install_gems

resources/puppetlabs/lein-ezbake/template/foss/ext/redhat/ezbake.spec.erb
melissa: Wanna add that to the commit message?

shrug: Done

cprice404:
:see_no_evil: if RE people are good with this, and if @shrug 's puppetserver test branch is working the way y'all would want the gems to end up being packaged, then I'm +1 on this one.

cprice404:
I haven't figured out how this new PR review stuff works on github yet, but I assume that when it says "@melissa approved these changes", that's a +1?  :)

@underscorgan anything else you wanted to look into on this one?

shrug:
Compare with https://github.com/puppetlabs/ezbake/pull/359

underscorgan:
@shrug I think I prefer this to #359, I feel like being deliberate about what extra dependencies we're adding rather than packaging up whatever's in a directory seems good?

I'm not totally sold on that opinion though (since I'm not sure I've fully wrapped my head around this), so feel free to convince me #359 is better!

scotje:
From an outside perspective, opt-in (this PR's approach) seems better. Are there any clear tradeoffs versus the approach of #359? We'd like to get one of these two merged so we can start moving forward on packaging gems with puppetserver.

shrug:
#359 is simpler and doesn't add any new functionality to ezbake, with the downside of adding java as a build-dependency across all of the projects, which doesn't seem out of line. 

I think I prefer #359 but I don't have a strong preference either way.

cprice404:
@shrug I probably prefer the simplicity of #359 too if you don't have concerns about java being universally required, but since this repo is just for building Clojure projects, that doesn't really seem like a big deal to me.

If that aligns with your thinking, then can we just go ahead and merge #359?

underscorgan:
Looking through these again with more ☕ in my system I think I prefer this one, but this is based on an assumption, so let me know if this is wrong:
* The need to be able to add java as a build dependency is blocking gem bundling with server

I feel like this is a more generic solution and like something we might run into in the future, with needing more build-time dependencies.

cprice404:
@underscorgan I think that your assumption is accurate.  FWIW I'm fine with this one over #359 if that's what RE prefers.  Would be awesome to get one of the two merged in the next day or two, though.

resources/puppetlabs/lein-ezbake/template/pe/ext/debian/control.erb
```diff
@@ -2,7 +2,10 @@ Source: <%= EZBake::Config[:project] %>
 Section: utils
 Priority: optional
 Maintainer: Puppet Labs <info@puppetlabs.com>
-Build-Depends: debhelper (>= 7.0.0~), cdbs, bc, mawk, lsb-release, puppet-agent
+Build-Depends: debhelper (>= 7.0.0~), cdbs, bc, mawk, lsb-release, puppet-agent<%=
+    if !EZBake::Config[:debian][:additional_build_dependencies].empty?
```
mckern: Um, isn't this just `unless EZBake::Config[:debian][:additional_build_dependencies].empty?`?

mckern: The way these are handled could stand to be refactored, but this will work. Can we make a note to clean up the way this list is constructed later on? Or is that something we can fix relatively quickly in a followup `(maint)` PR?

resources/puppetlabs/lein-ezbake/template/pe/ext/redhat/ezbake.spec.erb
```diff
@@ -103,6 +103,9 @@ Requires:         pe-puppet-enterprise-release
 # net-tools is required for netstat usage in service unit file
 # See: https://tickets.puppetlabs.com/browse/SERVER-338
 Requires:         net-tools
+<% EZBake::Config[:redhat][:additional_build_dependencies].each do |dep| %>
```
mckern: My kingdom for a helper method 😞

mckern:
YOLO

PR #360, opened 2016-09-19
(EZ-99) Introduce TK stop script and call from service stop

camlow325:
This PR adds a TK stop subcommand script, which just gets the pid of
the service process, calls the kill_pid function to kill it, and nukes
the associated service pidfile.  This PR also modifies all of the OS init
scripts / systemd service definitions to invoke the new TK stop
subcommand to stop a service.

The commit also inlines some fully-qualified uses of
/opt/puppetlabs/server/apps/<app_name> with references to
$INSTALL_DIR, where possible.

For the SLES 11 init script, the startproc command line was also fixed
up such that the process will actually be started correctly and some
excess log message logic was removed.

### DIFF
--- a/resources/puppetlabs/lein-ezbake/template/foss/ext/debian/ezbake.init.erb
+++ b/resources/puppetlabs/lein-ezbake/template/foss/ext/debian/ezbake.init.erb
@@ -63,7 +63,7 @@ do_start()
     <% end -%>
 
     start-stop-daemon --start --quiet --chuid $USER --oknodo --pidfile $PIDFILE --chdir $INSTALL_DIR \
-      --startas "/opt/puppetlabs/server/apps/${REALNAME}/bin/${REALNAME}" -- start >> /var/log/puppetlabs/${REALNAME}/${REALNAME}-daemon.log 2>&1
+      --startas "${INSTALL_DIR}/bin/${REALNAME}" -- start >> /var/log/puppetlabs/${REALNAME}/${REALNAME}-daemon.log 2>&1
     retval=$?
 
     <% if not EZBake::Config[:debian][:post_start_action].empty? -%>
@@ -82,16 +82,7 @@ do_start()
 #
 do_stop()
 {
-    # Return
-    #   0 if daemon has been stopped or was already stopped
-    #   2 if daemon could not be stopped
-    #   other if a failure occurred
-    start-stop-daemon --stop --quiet --oknodo --retry=TERM/${SERVICE_STOP_RETRIES}/KILL/5 --pidfile $PIDFILE --exec $JAVA_BIN
-    RETVAL="$?"
-    [ "$RETVAL" = 2 ] && return 2
-    # Many daemons don't delete their pidfiles when they exit.
-    rm -f $PIDFILE
-    return "$RETVAL"
+    "${INSTALL_DIR}/bin/${REALNAME}" stop
 }
 
 #
@@ -118,27 +109,24 @@ get_status_q()
 do_restart()
 {
     do_stop
-    case "$?" in
-        0|1)
-            do_start
-            case "$?" in
-                0) log_end_msg 0 ;;
-                1) log_end_msg 1 ;; # Old process is still running
-                *) log_end_msg 1 ;; # Failed to start
-            esac
-            ;;
-        *)
-            # Failed to stop
-            log_end_msg 1
-            ;;
-    esac
+    if [ "$?" -eq 0 ]; then
+        do_start
+        case "$?" in
+            0) log_end_msg 0 ;;
+            1) log_end_msg 1 ;; # Old process is still running
+            *) log_end_msg 1 ;; # Failed to start
+        esac
+    else
+        # Failed to stop
+        log_end_msg 1
+    fi
 }
 
 #
 # Function that sends a SIGHUP to the daemon/service
 #
 do_reload() {
-    /opt/puppetlabs/server/apps/${REALNAME}/bin/${REALNAME} reload
+    "${INSTALL_DIR}/bin/${REALNAME}" reload
     return $?
 }
 
@@ -157,10 +145,11 @@ case "$1" in
         [ "$VERBOSE" != no ] && log_daemon_msg "Stopping $DESC" "$NAME"
         do_stop
         RETVAL="$?"
-        case "$RETVAL" in
-            0|1) [ "$VERBOSE" != no ] && log_end_msg 0 ;;
-            2) [ "$VERBOSE" != no ] && log_end_msg 1 ;;
-        esac
+        if [ "$RETVAL" -eq 0 ]; then
+            [ "$VERBOSE" != no ] && log_end_msg 0
+        else
+            [ "$VERBOSE" != no ] && log_end_msg 1
+        fi
         exit "$RETVAL"
         ;;
     status)
--- a/resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
+++ b/resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
@@ -9,7 +9,12 @@ dir=$(dirname "$restartfile")
 realname="<%= EZBake::Config[:real_name] %>"
 PIDFILE="/var/run/puppetlabs/${realname}/${realname}.pid"
 
-[ -e "${INSTALL_DIR}/ezbake-functions.sh" ] && . "${INSTALL_DIR}/ezbake-functions.sh"
+if [ ! -e "${INSTALL_DIR}/ezbake-functions.sh" ]; then
+    echo "Unable to find ${INSTALL_DIR}/ezbake-functions.sh script, failing start." 1>&2
+    exit 1
+fi
+
+. "${INSTALL_DIR}/ezbake-functions.sh"
 
 write_pid_file() {
     if [ ! -d "/var/run/puppetlabs/${realname}" ]; then
--- a/resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
+++ b/resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
@@ -5,10 +5,11 @@ pid=$(pgrep -f <%= EZBake::Config[:uberjar_name] %>)
 
 restartfile=<%= EZBake::Config[:restart_file] %>
 start_timeout=<%= EZBake::Config[:start_timeout] %>
-stop_timeout=60
 dir=$(dirname "$restartfile")
 realname="<%= EZBake::Config[:real_name] %>"
 PIDFILE="/var/run/puppetlabs/${realname}/${realname}.pid"
+
+[ -e "${INSTALL_DIR}/ezbake-functions.sh" ] && . "${INSTALL_DIR}/ezbake-functions.sh"
 
 write_pid_file() {
     if [ ! -d "/var/run/puppetlabs/${realname}" ]; then
@@ -20,28 +21,7 @@ write_pid_file() {
 
 terminate_java_process() {
     echo "Startup script was terminated before completion" 1>&2
-    kill -TERM $pid >/dev/null 2>&1
-    sleep 0.1
-
-    timeout=$stop_timeout
-    kill -0 $pid >/dev/null 2>&1
-    while [ $? -eq 0 ] && [ $timeout -ne 0 ]; do
-        sleep 1
-        ((timeout--))
-        kill -0 $pid >/dev/null 2>&1
-    done
-
-    if [ $timeout -eq 0 ]; then
-        echo "Java process $pid not terminated gracefully after $stop_timeout seconds" 1>&2
-        kill -KILL $pid >/dev/null 2>&1
-        sleep 1
-        kill -0 $pid >/dev/null 2>&1
-        if [ $? -eq 0 ]; then
-            echo "Java process $pid not killed after SIGKILL" 1>&2
-        fi
-    fi
-
-    rm -f "$PIDFILE"
+    kill_pid $pid $PIDFILE $SERVICE_STOP_RETRIES
     exit 1
 }
 
--- a/resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
+++ b/resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
@@ -30,7 +30,7 @@ terminate_java_process() {
     exit 1
 }
 
-if [ ! -z $pid ]; then
+if [ -n "$pid" ]; then
     write_pid_file $pid
     exit 0
 fi
--- a/resources/puppetlabs/lein-ezbake/template/global/ext/cli/stop.erb
+++ b/resources/puppetlabs/lein-ezbake/template/global/ext/cli/stop.erb
@@ -5,7 +5,12 @@ pid=$(pgrep -f <%= EZBake::Config[:uberjar_name] %>)
 realname="<%= EZBake::Config[:real_name] %>"
 PIDFILE="/var/run/puppetlabs/${realname}/${realname}.pid"
 
-[ -e "${INSTALL_DIR}/ezbake-functions.sh" ] && . "${INSTALL_DIR}/ezbake-functions.sh"
+if [ ! -e "${INSTALL_DIR}/ezbake-functions.sh" ]; then
+    echo "Unable to find ${INSTALL_DIR}/ezbake-functions.sh script, failing stop." 1>&2
+    exit 1
+fi
+
+. "${INSTALL_DIR}/ezbake-functions.sh"
 
 if [ -z $pid ]; then
     rm -f "$PIDFILE"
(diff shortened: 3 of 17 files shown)


resources/puppetlabs/lein-ezbake/template/pe/ext/redhat/init.suse.erb
```diff
@@ -104,39 +92,20 @@ start() {
...
-    if [ ! -s "${PIDFILE}" ] ; then
-        print_service_pid > "${PIDFILE}"
-    fi
-
-    killproc -p "${PIDFILE}" -t"${SERVICE_STOP_RETRIES}"s -- "${JAVA_BIN}" "-XX:OnOutOfMemoryError=\"kill -9 %p\"" "${JAVA_ARGS}"
+    "${INSTALL_DIR}/bin/${realname}" stop
     rc_status -v
     retval=$?
 
-    if [ "$retval" = 0 ]; then
```
camlow325: Same rationale for the seemingly redundant status message line here as with the `start` command.

resources/puppetlabs/lein-ezbake/template/pe/ext/redhat/init.suse.erb
```diff
@@ -149,27 +118,23 @@ sl_status_q() {
...
 reload() {
-    echo -n $"Reloading ${prog}:"
-    /opt/puppetlabs/server/apps/${realname}/bin/${realname} reload
+    echo -n $"Reloading ${prog}: "
+    "${INSTALL_DIR}/bin/${realname}" reload
     rc_status -v
     retval=$?
 
-    if [ "$retval" = 0 ]; then
-        log_success_msg $"${prog} reloaded"
```
camlow325: Same rationale for the seemingly redundant status message line here as with the `start` command.

camlow325:
I added dfe63dd to cover a minor oversight where the service's pid file wasn't being removed in the event that the service dies during a reload attempt.

resources/puppetlabs/lein-ezbake/template/pe/ext/redhat/init.suse.erb
```diff
@@ -84,16 +75,13 @@ start() {
     # startproc will change users, so make sure that user has permission
     # to access the present working directory.
     cd "${INSTALL_DIR}"
-    startproc -u "${USER}" -l "${LOGFILE}" -p "${PIDFILE}" "/opt/puppetlabs/server/apps/${realname}/bin/${realname} start >> /var/log/puppetlabs/${realname}/${realname}-daemon.log 2>&1"
+    startproc -u "${USER}" -l "${LOGFILE}" -p "${PIDFILE}" -w -- "${INSTALL_DIR}/bin/${realname}" start >/dev/null 2>&1
```
camlow325: I changed the command line passed to `startproc` here because the original form did not work.  `startproc` would error out because it appeared to want an executalble file as the first argument after the `${PIDFILE}` one - not an argument composed of an executable and the other arguments to be passed to that executable.

The latest code on the stable branch formats each of those as separate arguments rather than a single one.  That approach worked fine for me in testing on SLES 11.

I hadn't been successful in getting the piped output from the start command to be appended into the daemon.log file.  It does appear that the target of the -l option, `"${LOGFILE}"`, receives output from the start command and launched Java process, though.  `${LOGFILE}` here is set to `/var/log/puppetlabs/puppetserver/puppetserver.log`.  It would seem more consistent with other non-systemd OSes if we were to use `/var/log/puppetlabs/puppetserver/puppetserver-daemon.log` instead here although it would arguably be a compatibility break for prior users of PE on SLES 11 to switch the log file target at this point.

I'd be curious to hear what others think here.  Worthwhile to switch the -l argument to the daemon.log file instead at this point?

camlow325:
@mckern, @underscorgan, @melissa, @shrug, @haus, @cprice404 - sorry to be a pest, but any chance I could get a couple of reviews on this sometime soon.  Thanks much!

resources/puppetlabs/lein-ezbake/template/foss/ext/debian/ezbake.init.erb
```diff
@@ -118,27 +109,24 @@ get_status_q()
 do_restart()
 {
     do_stop
-    case "$?" in
-        0|1)
```
cprice404: hmm, so, today, if we get an exit code of 1 on the stop, we just ignore it and try to start anyway?

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
```diff
@@ -5,11 +5,12 @@ pid=$(pgrep -f <%= EZBake::Config[:uberjar_name] %>)
 
 restartfile=<%= EZBake::Config[:restart_file] %>
 start_timeout=<%= EZBake::Config[:start_timeout] %>
-stop_timeout=60
 dir=$(dirname "$restartfile")
 realname="<%= EZBake::Config[:real_name] %>"
 PIDFILE="/var/run/puppetlabs/${realname}/${realname}.pid"
 
+[ -e "${INSTALL_DIR}/ezbake-functions.sh" ] && . "${INSTALL_DIR}/ezbake-functions.sh"
```
cprice404: If the functions script doesn't exist, then this script won't work because it references stuff from in there.  So is there any point in checking for the existence first?

In any case, I presume this line will cause a (desirable) failure because of the `set +e` if the script doesn't exist?

resources/puppetlabs/lein-ezbake/template/foss/ext/debian/ezbake.init.erb
haus: so the `--oknodo` flag makes me doubt we would ever get a non-zero return on the stop command.

resources/puppetlabs/lein-ezbake/template/pe/ext/redhat/init.suse.erb
cprice404: I would change it to the daemon log, personally.

cprice404:
Left a few minor questions, and also i have no idea what I'm doing, but this LGTM.

resources/puppetlabs/lein-ezbake/template/foss/ext/debian/ezbake.init.erb
camlow325: Sometimes, I think.  Today, if the `start-stop-daemon` fails with a retval of 2, 'end of schedule reached and processes were still running', the `restart` won't try to move on to the `start` and will return a `1` instead.  I suppose there could be other reasons that the stop might fail today where the start would still be attempted but I'm not coming up with any good ones.  With the new stop subcommand as it is on the PR, we wouldn't be able to distinguish the reasons a stop would fail (note that if the service had already been stopped, the result of the stop would be a success/0), so I just unconditionally have it not move on to the start.  Are you thinking something needs to be changed here?

haus: the `--oknodo` means return 0 if the service is stopped fwiw.

resources/puppetlabs/lein-ezbake/template/pe/ext/redhat/init.suse.erb
camlow325: Cool, that's what I'd like to do as well.  Curious if @haus or @shrug in particular has a strong opinion on this point.

resources/puppetlabs/lein-ezbake/template/foss/ext/debian/ezbake.init.erb
cprice404: @camlow325 no, just asking.

resources/puppetlabs/lein-ezbake/template/pe/ext/redhat/init.suse.erb
```diff
@@ -84,16 +75,13 @@ start() {
...
+    [ ! -z $(pgrep -f <%= EZBake::Config[:uberjar_name] %>) ]
```
haus: that seems reasonable.

i think you may want `-n` here instead of `! -z` for readability and I'd recommend putting the subshell in "" to ensure that a value is here, even if it is the empty string.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
camlow325: Actually, I think I'd need a `set -e` to have it fail if the script doesn't exist.  I had gone with the `set +e` at the top - which is probably redundant with the default behavior - because some of the code in `kill_pid` relies upon handling non-zero exit codes, where we don't want the script to fail, but was trying to avoid blocking `set -e` / `set +e` calls throughout, like @haus had suggested on another recent PR.

So, yeah, as it stands I guess the `-e` check for the file's existence is a no-op.  I could change it to error out with a specific error message if that makes more sense or just assume the script is there and source it - yolo.  Wasn't sure how much error handling for this was useful to include here.  Suppose the one error message wouldn't be too onerous to look at?

resources/puppetlabs/lein-ezbake/template/pe/ext/redhat/init.suse.erb
haus: do it. do it now.

resources/puppetlabs/lein-ezbake/template/foss/ext/debian/ezbake.init.erb
haus: what does the stop script return if the service is already stopped?

camlow325: It returns 0, so I think we'd be backward-compatible here for that case, right?

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
haus: no, `set +e` means don't fail on errors. `set -e` means fail on errors.

this line will not cause a failure if ezbake-functions.sh doesn't exist.

resources/puppetlabs/lein-ezbake/template/pe/ext/redhat/init.suse.erb
camlow325: Ah, good calls on both.  Will do!

resources/puppetlabs/lein-ezbake/template/global/ext/cli/stop.erb
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+set +e
+
+pid=$(pgrep -f <%= EZBake::Config[:uberjar_name] %>)
```
haus: we should make sure we pull in whatever supplies pgrep in the package dependencies. It should be there on most systems already, but it won't hurt.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/stop.erb
```diff
@@ -0,0 +1,15 @@
...
+realname="<%= EZBake::Config[:real_name] %>"
+PIDFILE="/var/run/puppetlabs/${realname}/${realname}.pid"
+
+[ -e "${INSTALL_DIR}/ezbake-functions.sh" ] && . "${INSTALL_DIR}/ezbake-functions.sh"
+
+if [ -z $pid ]; then
```
haus: recommend wrapping `$pid` in double quotes.

resources/puppetlabs/lein-ezbake/template/global/ext/ezbake-functions.sh.erb
```diff
@@ -61,6 +61,57 @@ wait_for_pidfile()
...
+        kill -0 $pid >/dev/null 2>&1
+        if [ $? -eq 0 ]; then
+            echo "Process $pid not killed after SIGKILL" 1>&2
+            return 1
+        else
+            echo "Process $pid killed after SIGKILL" 1>&2
+        fi
+    fi
+
+    if [ ! -z $pidfile ]; then
```
haus: same comment about wrapping `$pidfile` in double quotes.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
camlow325: @haus - you have any strong opinion as to whether we check for the existence of the file or not?  If we do want to check for it, better to wrap a `set -x` and `set +x` around it or echo a specific message and error out.  I'm kind of leaning toward the latter, I guess, but don't care too much.

resources/puppetlabs/lein-ezbake/template/foss/ext/debian/ezbake.init.erb
haus: yea, that's perfect.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
haus: i think `set -e` errors are super opaque in general, but with how it is currently we wouldn't see the failure until we tried to use the functions later on, and i forget how that would be surfaced, but i doubt it would be helpful. probably better to have an explicit error + exit if the functions file doesn't exist.

cprice404: damn it.  I can never remember which meaning any of those four characters (+,-,e,x) have in bash. :/

mckern: I would not use `set -e` in an init script. It's a pretty brittle setting that is best used when developing/debugging a script -- in practice it should handle errors more gracefully than simply failing entirely.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/stop.erb
camlow325: I did a survey across a few OSes - CentOS 6/7, Ubuntu 12.04/14.04/16.04, SLES 11/12, Debian 8 - and for each, I saw that `procps` was providing pgrep.  It appears we have already been using `pgrep` for RedHat derivatives in shipping init scripts but nothing for Debian, AFAICT.

I agree with the suggestion - just want to make sure I understand what you're asking me to do here.  Would I add `procps` as a `Requires` in the "ezbake.spec.erb" files for rpm and as a `Depends` in the control.erb for deb - for both foss and pe?  Anywhere else?

haus: i think that's the right thing yea.

camlow325: Hmm, I noticed that for CentOS 7, the full package name is `procps-ng` whereas it is only `procps` for CentOS 6.  For a package I built for CentOS 7 where the dependency was expressed as `procps`, it did appear that the dependency was properly satisfied, though.  Would it be best to run with the shorter name to keep it the same for all RedHat derivatives or should we somehow be more specific per OS?

resources/puppetlabs/lein-ezbake/template/pe/ext/redhat/init.suse.erb
camlow325: Did this in 7f149a5.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
camlow325: Added the explicit error messages in 22a6b54.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/stop.erb
camlow325: @haus I went ahead and just added `procps` (not conditioning for `procps-ng` where available) in 43d35b0.  Worked fine for CentOS 7 despite it having `procps-ng` instead.  I tested this on some other distros and it seemed fine - CentOS 6, Debian 8, Ubuntu 12.04, Ubuntu 16.04, SLES 11/12.  If it proves to be a problem, we could make it conditional later.  Sound okay?

resources/puppetlabs/lein-ezbake/template/global/ext/cli/stop.erb
camlow325: Did this in 4065895.

resources/puppetlabs/lein-ezbake/template/global/ext/ezbake-functions.sh.erb
camlow325: Did this in 4065895.

camlow325:
@cprice404, @haus - I think I've hit all of your feedback with follow-up commits.

RE folks - @shrug, @melissa, @underscorgan, @mckern - has at least 1 of you reviewed this enough to feel comfortable with this being merged?

camlow325:
Rebased on HEAD of the master branch, 79dc056.

camlow325:
@cprice404, @haus - are you 👍 on this now?

@shrug, I would very much like to have someone from your team do a full review of this before it comes in but if no-one is going to be available in the near-future are you okay on the review from @cprice404 and @haus to just let this come in?  This is blocking a couple of other PRs that I'd like to get through soon, if possible.  Thanks!

shrug:
👍 
@haus or @cprice404 did you have anything else or are we good to merge?

cprice404:
+1 lessdoit!

PR #361, opened 2016-09-20
(EZ-100) Add timeout settings for reload and stop

camlow325:
This PR adds optional Ezbake config timeout settings for reload and stop, defaulting to 2 minutes and 60 seconds, respectively.  The settings are attached to the new `RELOAD_TIMEOUT` and pre-existing `SERVICE_STOP_TIMEOUT` default/sysconfig settings.

The PR also includes some code to terminate the Java process if the start timeout is reached during a startup attempt.

### DIFF
--- a/resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
+++ b/resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
@@ -2,7 +2,8 @@
 set +e
 
 restartfile=<%= EZBake::Config[:restart_file] %>
-timeout=<%= EZBake::Config[:start_timeout] %>
+reload_timeout=${RELOAD_TIMEOUT:-<%= EZBake::Config[:reload_timeout] %>}
+timeout=$reload_timeout
 realname="<%= EZBake::Config[:real_name] %>"
 PIDFILE="/var/run/puppetlabs/${realname}/${realname}.pid"
 
@@ -13,6 +14,7 @@ if [ -r "$restartfile" ];  then
         echo "Service not running so cannot be reloaded" 1>&2
         exit 1
     fi
+    sleep 0.1
     cur=$(head -n 1 "$restartfile")
     initial=$cur
     while [ "$cur" == "$initial" ] ;do
@@ -22,12 +24,12 @@ if [ -r "$restartfile" ];  then
             rm -f "$PIDFILE"
             exit 1
         fi
-        sleep 0.1
+        sleep 1
         cur=$(head -n 1 "$restartfile")
 
         ((timeout--))
         if [ $timeout -eq 0 ]; then
-            echo "Reload timed out after <%= EZBake::Config[:start_timeout].to_i / 10 %> seconds"
+            echo "Reload timed out after $reload_timeout seconds"
             exit 1
         fi
     done
--- a/resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
+++ b/resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
@@ -1,16 +1,16 @@
 #!/usr/bin/env bash
 set +e
 
-restartfile=<%= EZBake::Config[:restart_file] %>
-reload_timeout=${RELOAD_TIMEOUT:-<%= EZBake::Config[:reload_timeout] %>}
-timeout=$reload_timeout
+restartfile="<%= EZBake::Config[:restart_file] %>"
+reload_timeout="${RELOAD_TIMEOUT:-<%= EZBake::Config[:reload_timeout] %>}"
+timeout="$reload_timeout"
 realname="<%= EZBake::Config[:real_name] %>"
 PIDFILE="/var/run/puppetlabs/${realname}/${realname}.pid"
 
 if [ -r "$restartfile" ];  then
-    cur=$(head -n 1 "$restartfile")
-    initial=$cur
-    pid=$(pgrep -f <%= EZBake::Config[:uberjar_name] %>)
+    cur="$(head -n 1 "$restartfile")"
+    initial="$cur"
+    pid="$(pgrep -f <%= EZBake::Config[:uberjar_name] %>)"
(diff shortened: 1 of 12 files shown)


camlow325:
This is based on #360.  Only the last 3 commits are new for this PR.  Once the other PR lands, I'll rebase this PR off of the master branch again and, then, this should be ready for merge.

camlow325:
@shrug, @underscorgan, @melissa, @mckern, @cprice404, @haus - this is a PR that builds on #360 that I'd like to have at least a subset of you review but definitely would like to get reviews in for #360 first.  Thanks much!

camlow325:
This has been rebased on the HEAD of the master branch at 0aa6e76 now that #360 has been merged and should now be ready for merge.

@shrug, @melissa, @haus, @cprice404 - could some of you take a look as soon as you get a chance?

resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
```diff
@@ -13,6 +14,7 @@ if [ -r "$restartfile" ];  then
         echo "Service not running so cannot be reloaded" 1>&2
         exit 1
     fi
+    sleep 0.1
```
cprice404: What is this sleep for?  Not that it's a problem, just trying to grok.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
```diff
@@ -13,6 +14,7 @@ if [ -r "$restartfile" ];  then
...
     cur=$(head -n 1 "$restartfile")
     initial=$cur
```
cprice404: Should we be grabbing the value for `initial` *before* we send the HUP signal?

cprice404:
:see_no_evil: just had a couple of minor questions.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
camlow325: It's arbitrary.  It makes the assumption that it may take the application a little bit of time - but not much -- to process the HUP signal and reload in cases where there's no work in process that has to be gracefully stopped.  Just seemed like for cases where the reload is fairly quick that we'd have a better chance of being able to detect it has finished here before going into the 1 second polling loop.  I took this idea from the killproc implementation from the init function in some OS distributions about how they wait for a process to stop - quick sleep up front (.1 second) followed by the 1 second sleeps later if it doesn't complete right away.

All that said, I'm good with removing this line.  I don't care all that much.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
camlow325: Ah, good call.  Hadn't noticed that previously.  I'll change that.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
cprice404: It's ok with me if you think it might help.  Was just curious.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
camlow325: I switched the order for grabbing the `initial` value to before sending the HUP signal in c46b39da.

cprice404:
+1.  over to you, @shrug / @melissa / @underscorgan .

resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
```diff
@@ -2,32 +2,34 @@
...
 restartfile=<%= EZBake::Config[:restart_file] %>
-timeout=<%= EZBake::Config[:start_timeout] %>
+reload_timeout=${RELOAD_TIMEOUT:-<%= EZBake::Config[:reload_timeout] %>}
+timeout=$reload_timeout
 realname="<%= EZBake::Config[:real_name] %>"
 PIDFILE="/var/run/puppetlabs/${realname}/${realname}.pid"
 
 if [ -r "$restartfile" ];  then
+    cur=$(head -n 1 "$restartfile")
+    initial=$cur
```
shrug: Nitpick: we should probably quote the variables

resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
```diff
@@ -2,32 +2,34 @@
 set +e
 
 restartfile=<%= EZBake::Config[:restart_file] %>
-timeout=<%= EZBake::Config[:start_timeout] %>
+reload_timeout=${RELOAD_TIMEOUT:-<%= EZBake::Config[:reload_timeout] %>}
+timeout=$reload_timeout
```
shrug: Nitpick: we should probably quote the variables

camlow325: This does an `-eq` expression below so if the `timeout` were ever anything that doesn't coerce to an integer expression, the `-eq` check would blow up anyway.  Given that, I was a bit apathetic about quoting the timeout.  Wouldn't seem to cause any harm, though, so I can do that.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
camlow325: Will do, thanks.  New commit forthcoming...

camlow325: @shrug - I pushed up b5d48f5, which adds in some extra quoting around variables in the start, stop, and reload scripts.  Let me know if you think that captures the ones you were concerned about.  Thanks!

camlow325:
Thanks, @shrug.  Since we have two 👍s now, do you think any one else on your team needs to review this before merging?

PR #362, opened 2016-09-20
(EZ-98) Fail an ezbake package build attempt if restart-file not set

camlow325:
This commit causes an ezbake package build attempt to fail if the
`restart-file` setting is not specified in the ezbake config.  This
check is being added because `restart-file` is essentially a required
setting for Ezbake 1.0+ in that it would be impossible to determine when
a service has been completely started without the use of the setting.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -4,11 +4,26 @@
   :license {:name "Apache License 2.0"
             :url "http://www.apache.org/licenses/LICENSE-2.0"}
 
-  :dependencies [[me.raynes/fs "1.4.6" :exclusions [org.clojure/clojure]]
+  :dependencies [;; begin version conflict resolution dependencies
+                 [org.clojure/tools.reader "1.0.0-beta1"]
+                 ;; end version conflict resolution dependencies
+
+                 [me.raynes/fs "1.4.6" :exclusions [org.clojure/clojure]]
                  [me.raynes/conch "0.8.0"]
                  [clj-time "0.6.0"]
                  [prismatic/schema "1.0.4"]
-                 [puppetlabs/typesafe-config "0.1.3" :exclusions [org.clojure/clojure]]]
+
+                 [puppetlabs/typesafe-config "0.1.5" :exclusions [org.clojure/clojure]]
+
+                 ;; trapperkeeper pulls in core.cache via core.async.  Since
+                 ;; lein pulls in its own (older) version of core.cache,
+                 ;; running ezbake as a plugin to another project via lein
+                 ;; produces a "Could not locate clojure/data/priority_map__init.class"
+                 ;; error.  Seems related to
+                 ;; https://github.com/technomancy/leiningen/issues/1563.  Excluding
+                 ;; core.cache here appears to avoid the conflict without requiring
+                 ;; consuming projects to do the same.
+                 [puppetlabs/trapperkeeper "1.5.0" :exclusions [org.clojure/core.cache]]]
 
   :repositories [["releases" "http://nexus.delivery.puppetlabs.net/content/repositories/releases/"]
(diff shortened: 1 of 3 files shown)


shrug:
👍 from RE for the intent and it generally looks fine, someone better at Clojuring might want to take a look.

camlow325:
Cool, thx, @shrug.  @haus or @cprice404 - could one of you check this one out?  Pretty simple compared to some of my other recent ones.

haus:
@camlow325 seems fine. if we have more required settings, we might want a more generic solution that we just wrap `make-template-map` (like schema maybe, with required-keys for the things we care about).

Do we care about validating that the restart-file is also set in the :global section of the tk config?

camlow325:
@haus, thanks for the suggestions.  I agree with both and pushed up a new commit in place of the previous one, 38a19c9, which attempts to address them.

project.clj
```diff
@@ -4,11 +4,26 @@
...
+
+                 ;; trapperkeeper pulls in core.cache via core.async.  Since
+                 ;; lein pulls in its own (older) version of core.cache,
+                 ;; running ezbake as a plugin to another project via lein
+                 ;; produces a "Could not locate clojure/data/priority_map__init.class"
+                 ;; error.  Seems related to
+                 ;; https://github.com/technomancy/leiningen/issues/1563.  Excluding
+                 ;; core.cache here appears to avoid the conflict without requiring
+                 ;; consuming projects to do the same.
+                 [puppetlabs/trapperkeeper "1.5.0" :exclusions [org.clojure/core.cache]]]
```
camlow325: I really don't like having to do this but couldn't figure out another way to work around this.  If others -- maybe @cprice404 -- have a suggestion as to an alternate way to handle this dependency conflict, I'm definitely open to changing this.

src/leiningen/ezbake.clj
```diff
@@ -26,6 +26,7 @@ Actions:
                   ezbake-core/resource-path)]
         (lein-ezbake-core/prepare-resource-dir project)
         (try
+          (ezbake-core/validate! project)
```
camlow325: @haus had suggested doing the validation of the `vars` from the project config via a parameter to the `make-template-map` function.  That would seem like a reasonable way to go but I thought it might be even better to just have an explicit validation step before any of the package building process has started -- setup of the `./target` directory, etc. -- where any invalid config could cause the package build to fail.

haus:
@camlow325 looks great!

project.clj
cprice404: adding an explicit dependency on clojure and/or org.clojure/core.cache won't work?

cprice404:
LGTM other than the question about the `:exclusions` thing, and if what I suggested doesn't work I'm fine with it like this.

camlow325:
After talking with @cprice404 about this a bit offline, we decided to go a different way with the `restart-file` feature.  In [TK-405](https://tickets.puppetlabs.com/browse/TK-405), we'll be plumbing through a way to set the `restart-file` value via a TK command line argument.  In [EZ-101](https://tickets.puppetlabs.com/browse/EZ-101), we'll be changing EzBake to populate the value passed to the TK restart-file option.  This allows us the ability to use a default value for restart-file during EzBake package creation.  By doing all of this, we should be able to eliminate the condition where the `restart-file` setting doesn't have a value - making this PR partially moot.

I do think there's still some value in adding the schema validation for EzBacke config options but I think it would be best at this point to close this PR and put up a replacement which has just the extra schema validation and nothing that makes `restart-file` a required setting.  So ... closing.

camlow325:
I submitted the schema validation work - without any of the `restart-file` logic since that no longer exists as an ezbake config option - in a follow-up PR, #367.

PR #364, opened 2016-09-26
Debian systems can be runtime-switched between sysvinit and systemd

zeha:
d2ad06508bed added systemd support for Ubuntu xenial, and as noted there did not enable it for jessie. Also, Debian-based systems can be runtime-switched between systemd and sysvinit, so packages should always install support for both init systems.

Actually doing this makes the code smaller, and now supports jessie, too.

### DIFF
--- a/resources/puppetlabs/lein-ezbake/template/foss/ext/debian/postinst.erb
+++ b/resources/puppetlabs/lein-ezbake/template/foss/ext/debian/postinst.erb
@@ -4,12 +4,8 @@
 
 # On upgrade, we should restart the service if it's running
 if [ $1 = 'configure' -a -n $2 ] ; then
-  if [ -e "/lib/systemd/system/<%= EZBake::Config[:project] -%>.service" ] ; then
+  if [ -d /run/systemd/system ]; then
     # Using systemd
-    if [ -e '/etc/init.d/<%= EZBake::Config[:project] %>' ] ; then
-      # We need to clean up the old sysv files if they still exist
-      rm '/etc/init.d/<%= EZBake::Config[:project] %>'
-    fi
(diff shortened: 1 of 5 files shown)


zeha:
Notes:
- I did not really test this (sorry).
- The systemd check is what the debian build tools embed in the prerm/postinst scripts, so should be fine.
- This is related to https://tickets.puppetlabs.com/browse/SERVER-1564

cprice404:
@melissa just checking to see if you saw this since we were talking about this issue in chat the other day.

melissa:
@cprice404 yes, we have a ticket for this in the next sprint and will address it then

cprice404:
@melissa awesome, thanks!

mckern:
I am tacitly OK with this, but I'd like to see it tested before we merge it.

underscorgan:
@haus was there something weird with logctl not working if sysvinit and systemd scripts were both installed?

haus:
@underscorgan not sure. i don't remember even trying it.

melissa:
@zeha I pulled in some concepts from this PR and added a few updates that were left out. I'm still making my way through testing, but I think we should be able to get this to work. Thoughts?

zeha:
@melissa I had a quick look at 952c10838e7f - personally I'd take the opportunity to remove the hardcoded release names list and merge the two deb_xxx tasks, but if not doing this makes testing easier it should be alright?

melissa:
@zeha the way this code is being used, I think it's best for us to keep that list. I do agree that it's not the best to keep hardcoded lists like that around.

I've updated my PR to pull in your commit, and I've added additional commits with my changes. As such, I'm closing this PR in favor of https://github.com/puppetlabs/ezbake/pull/382

PR #367, opened 2016-09-29
(MAINT) Schema validate ezbake config before staging package

camlow325:
This commit adds some schema validation of the ezbake project config
before starting the package staging phase.


camlow325:
I pulled this over from a subset of the work in #362 that we decided not to merge.  This PR isn't holding anything up.  Can be merged whenever is convenient.

PR #368, opened 2016-09-29
(MAINT) Set umask to 027 in cli app script

camlow325:
In EZ-83, work was done across various init scripts / system service
definitions to set the default umask to 027.  Since some of the cli
subcommands - notably, foreground - are not typically initiated from an
init script / service context, they would still run with the default
system umask.  This commit adds the same explicit umask to the cli app
script for consistency - ensuring that when adhoc commands like
foreground are run that files like logs are not created with undesirable
world-accessible permissions.


camlow325:
@shrug - this one isn't blocking anything on our side but thought I'd put this up before I forgot about it.  Feel free to review whenever you have time.

PR #370, opened 2016-09-30
(SERVER-1412) Allow build scripts to be included in staging tarball

cprice404:
This commit adds support for creating an `ext/build-scripts/` dir
in a project's jar.  If found, ezbake will copy all of the files
found in this directory into the staging directory, so that they
will be included in the tarball that is sent up to the packaging
system.

These files will *not* be included in the final packages, but they
can be referenced during the package build, e.g. during the `install`
phase, in order to reduce duplication of logic in the ezbake config
files.

### DIFF
--- a/src/puppetlabs/ezbake/core.clj
+++ b/src/puppetlabs/ezbake/core.clj
@@ -32,6 +32,7 @@
 (def system-config-dir "ext/system-config/")
 (def shared-cli-apps-prefix "ext/cli/")
 (def docs-prefix "ext/docs/")
+(def build-scripts-prefix "ext/build-scripts/")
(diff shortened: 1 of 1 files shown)


cprice404:
@camlow325 @shrug what do you guys think of this?  I'm trying to come up with a way to reduce the duplication of all the gem install commands for EZ-89.

cprice404:
This allows me to, in the `install` section, do something like "sh ./ext/build-scripts/foo.sh".

camlow325:
Seems reasonable to me.  Didn't do any functional testing of it but 👍.

cprice404:
@shrug WDYT?

shrug:
👍 works for me

PR #373, opened 2016-10-10
(SERVER-1591) Handle managed dependencies

cprice404:
Prior to this commit, there were some places in ezbake where it
does things to resolve maven dependencies.  The relevant code
did not take into account the new `:managed-dependencies`
capabilities in lein 2.7.  This commit adds in the appropriate
handling of those dependencies where necessary.

### DIFF
--- a/src/puppetlabs/ezbake/dependency_utils.clj
+++ b/src/puppetlabs/ezbake/dependency_utils.clj
@@ -4,7 +4,8 @@
   (:require [cemerick.pomegranate.aether :as aether]
             [me.raynes.fs :as fs]
             [clojure.string :as str]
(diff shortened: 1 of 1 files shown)


cprice404:
@camlow325 I need to file a PR for puppetserver too.  That combined with this should get us building again.

camlow325:
🙈

camlow325:
I did a smoke test of this against a build of OSS puppetserver from [this PR](https://github.com/puppetlabs/puppetserver/pull/1235).  The ezbake build was successful and the beaker tests I ran against the package which do a couple of agent runs against the server passed.  I think this is good to merge then.

:shipit:

PR #375, opened 2016-10-12
(EZ-103) Address incorrect rundir perms

melissa:
Prior to this commit, we had been creating the rundir, but with
incorrect permissions. This wasn't an issue for sysv systems, but it's
starting to crop up at a problem for systemd systems. This commit
ensures the rundir is created with the correct directory owner.

### DIFF
--- a/resources/puppetlabs/lein-ezbake/template/foss/ext/debian/ezbake.init.erb
+++ b/resources/puppetlabs/lein-ezbake/template/foss/ext/debian/ezbake.init.erb
@@ -32,6 +32,8 @@ fi
 PATH=/sbin:/usr/sbin:/bin:/usr/bin
 NAME=<%= EZBake::Config[:project] %>
 REALNAME=<%= EZBake::Config[:real_name] %>
+USER=<%= EZBake::Config[:user] %>
+GROUP=<%= EZBake::Config[:group] %>
 DESC="<%= EZBake::Config[:project] %> Puppet Labs version-checking backend"
 JARFILE="<%= EZBake::Config[:uberjar_name] %>"
 PIDFILE=/var/run/puppetlabs/${REALNAME}/${REALNAME}.pid
@@ -61,6 +63,8 @@ do_start()
     <% EZBake::Config[:debian][:pre_start_action].each do |action| -%>
     <%= action %>
     <% end -%>
+
+    /usr/bin/install --directory --owner=${USER} --group=${GROUP} --mode=775 /var/run/puppetlabs/${REALNAME}
 
     start-stop-daemon --start --quiet --chuid $USER --oknodo --pidfile $PIDFILE --chdir $INSTALL_DIR \
       --startas "${INSTALL_DIR}/bin/${REALNAME}" -- start >> /var/log/puppetlabs/${REALNAME}/${REALNAME}-daemon.log 2>&1
--- a/resources/puppetlabs/lein-ezbake/template/foss/ext/redhat/init.erb
+++ b/resources/puppetlabs/lein-ezbake/template/foss/ext/redhat/init.erb
@@ -48,10 +48,7 @@ START_TIMEOUT=${START_TIMEOUT:-<%= EZBake::Config[:start_timeout] %>}
 
 find_my_pid() {
     pid=`pgrep -f "${JARFILE}"`
-    if [ ! -d  "/var/run/puppetlabs/${realname}" ] ; then
-        mkdir -p /var/run/puppetlabs/${realname}
-        chown -R $USER:$GROUP /var/run/puppetlabs/${realname}
-    fi
+    /usr/bin/install --directory --owner=${GROUP} --group=${GROUP} --mode=775 /var/run/puppetlabs/${realname}
     [ -n "$pid" ] && echo $pid > $PIDFILE
 }
 
--- a/resources/puppetlabs/lein-ezbake/template/global/install.sh.erb
+++ b/resources/puppetlabs/lein-ezbake/template/global/install.sh.erb
@@ -173,7 +173,7 @@ function task_install {
     ln -s "../apps/${real_name}/bin/<%= basename %>" "${DESTDIR}${symbindir}/<%= basename %>"
     ln -s "../server/apps/${real_name}/bin/<%= basename %>" "${DESTDIR}${uxbindir}/<%= basename %>"
 <% end -%>
-    install -d -m 0755 "${DESTDIR}${rundir}"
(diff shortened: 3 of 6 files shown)


melissa:
p.s., I haven't tested this yet.

resources/puppetlabs/lein-ezbake/template/foss/ext/debian/ezbake.init.erb
```diff
@@ -62,6 +64,11 @@ do_start()
     <%= action %>
     <% end -%>
 
+    if [ ! -d "/var/run/puppetlabs/${REALNAME}" ] ; then
```
camlow325: FWIW, in the ezbake.conf file in Puppet Server, we had pre-start-actions which did this:

```
/usr/bin/install --directory --owner={{user}} --group={{user}} --mode=775 /var/run/puppetlabs/puppetserve
```
The one advantage of doing it this way is that the permissions would be corrected even if the directory already existed at the time this code were run.

I'm open to doing it this way instead, though, if you think that's better.

camlow325: I believe `install` would just effectively do nothing if the directory already exists with the intended permissions.  Only reason I'm throwing that out is that we've been doing it this way via the pre-start-action in Puppet Server for a number of releases without any problems and it does seem a little more robust for the case that the directory might already exist with the wrong permissions at start time.  I could go either way on that, though.  @haus - do you have a strong opinion on this?

haus: no strong feelings. if the directory exists, hopefully we created it and chowned it.

camlow325: Only case I can think of that might be a problem maybe is if the user ran the start or foreground script and the permissions got set to root instead of the service user as a result or something.  If you want to just go with what you've got already, though, I'm A-ok with that.

resources/puppetlabs/lein-ezbake/template/global/install.sh.erb
```diff
@@ -173,7 +173,7 @@ function task_install {
     ln -s "../apps/${real_name}/bin/<%= basename %>" "${DESTDIR}${symbindir}/<%= basename %>"
     ln -s "../server/apps/${real_name}/bin/<%= basename %>" "${DESTDIR}${uxbindir}/<%= basename %>"
 <% end -%>
-    install -d -m 0755 "${DESTDIR}${rundir}"
+    install -d -m 0755 -o <%= EZBake::Config[:user] %> -g <%= EZBake::Config[:group] %> "${DESTDIR}${rundir}"
```
camlow325: Hmm, this appears to have caused a package build failure for both RedHat and Debian for me.  From a build of pe-puppet-server-extensions at http://jenkins-release.delivery.puppetlabs.net/job/pe-puppetdb-packaging-2016-10-12-21-48-52-4.3.0.SNAPSHOT.2016.10.12T2148 for a Debian platform, I saw:

~~~~
+++ install -d -m 0755 -o pe-puppetdb -g pe-puppetdb /tmp/buildd/pe-puppetdb-4.3.0/debian/pe-puppetdb/var/run/puppetlabs/puppetdb
install: invalid user 'pe-puppetdb'
make: *** [install/pe-puppetdb] Error 1
~~~~

I suppose this makes sense if we don't have fixed uids/gids for the user/group and those are determined instead when the useradd/groupadd command is run at package install time.

If we're resetting the permissions at service start time / tmpfiles.d, I'm wondering now if we'll even need this anyway.

camlow325: I think if we take this out though that, at least on Debian, we'll need to have the logic that creates/chowns the directory in the init script be able to chown the directory if it already exists - i.e., if it was installed with root ownership.  In that case, going with {{/usr/bin/install}} would seem like a better approach.

I put up an alternative PR for this, #376, which switches the mkdir/chown pattern over to {{/usr/bin/install}} in all places.  It also adds in some extra error handling for the start script which should cause the service to fail right away in the event that permissions on the directory cannot be corrected before the process is started.  I'll do some testing with that across a few OSes to see that it does what I expect it to do.

If there's an alternate way to get the run dir built into the package with the proper USER/GROUP for Debian like we apparently have for RedHat, I'm definitely open to adding that in as well.  Hopefully all of the other changes with whatever we merge will compensate for the lack of that in any event.

camlow325: This is the line that makes me think we at least set rundir ownership at install time to the proper USER/GROUP on RedHat: https://github.com/puppetlabs/ezbake/blob/1.1.1/resources/puppetlabs/lein-ezbake/template/foss/ext/redhat/ezbake.spec.erb#L271.  Not sure if it's possible to do something similar for Debian?

resources/puppetlabs/lein-ezbake/template/foss/ext/redhat/init.erb
```diff
@@ -48,10 +48,7 @@ START_TIMEOUT=${START_TIMEOUT:-<%= EZBake::Config[:start_timeout] %>}
 
 find_my_pid() {
     pid=`pgrep -f "${JARFILE}"`
-    if [ ! -d  "/var/run/puppetlabs/${realname}" ] ; then
-        mkdir -p /var/run/puppetlabs/${realname}
-        chown -R $USER:$GROUP /var/run/puppetlabs/${realname}
-    fi
+    /usr/bin/install --directory --owner=${GROUP} --group=${GROUP} --mode=775 /var/run/puppetlabs/${realname}
```
melissa: @haus @camlow325 I've also updated this line for rpm packages to maintain consistency

resources/puppetlabs/lein-ezbake/template/global/install.sh.erb
melissa: I was trying to find something similar for debian, but I'm not coming up with anything concrete. The closes thing debian has is to be able to specify directories (https://www.debian.org/doc/manuals/maint-guide/dother.en.html#dirs), but I don't see where we would be able to specify permissions

melissa: We might also be able to modify the install.sh script to ensure the user is created prior to the service directory setup, in which case we could insure correct ownership from installation for all users, not just those who are installing from a package

melissa: We might also be able to modify the install.sh script to ensure the user is created prior to the service directory setup, in which case we could insure correct ownership from installation for all users, not just those who are installing from a package

melissa:
Closing in favor of https://github.com/puppetlabs/ezbake/pull/376

PR #376, opened 2016-10-13
(EZ-103) Address incorrect rundir perms

camlow325:
Prior to this commit, we had been creating the rundir, but sometimes
with permissions which prevented the service from being able to create
its own pidfile.  In these cases, the Java process could be launched
successfully and the service start command could still return success.
Any subsequent attempts to ask the service framework for status on the
service, however, would return failure (not started).

This commit should ensure that the rundir is always created with
permissions which allow the service to create its own pidfile.  This
commit also adds some error handling to return failure at service
startup if the pidfile unexpectedly could not be written for some
reason.

### DIFF
--- a/resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
+++ b/resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
@@ -7,21 +7,29 @@ restartfile="/opt/puppetlabs/server/data/<%= EZBake::Config[:real_name] %>/resta
 start_timeout="${START_TIMEOUT:-<%= EZBake::Config[:start_timeout] %>}"
 
 realname="<%= EZBake::Config[:real_name] %>"
-PIDFILE="/var/run/puppetlabs/${realname}/${realname}.pid"
+rundir="/var/run/puppetlabs/${realname}"
+PIDFILE="${rundir}/${realname}.pid"
 
 if [ ! -e "${INSTALL_DIR}/ezbake-functions.sh" ]; then
     echo "Unable to find ${INSTALL_DIR}/ezbake-functions.sh script, failing start." 1>&2
     exit 1
 fi
 
+/usr/bin/install --directory --owner=$USER --group=$GROUP --mode=755 "$rundir"
+if [ $? -ne 0 ]; then
+    echo "Unable to create/set permissions for rundir: ${rundir}" 1>&2
+    exit 1
+fi
+
 . "${INSTALL_DIR}/ezbake-functions.sh"
 
 write_pid_file() {
-    if [ ! -d "/var/run/puppetlabs/${realname}" ]; then
-        mkdir -p /var/run/puppetlabs/${realname}
-        chown -R $USER:$GROUP /var/run/puppetlabs/${realname}
+    echo "$pid" > "$PIDFILE"
+    if [ $? -ne 0 ]; then
+        echo "Unable to write pid file: ${PIDFILE}" 1>&2
+        terminate_java_process
+        exit 1
     fi
-    echo $1 > $PIDFILE
 }
 
 terminate_java_process() {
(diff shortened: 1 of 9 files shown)


resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
```diff
@@ -7,21 +7,29 @@ restartfile="/opt/puppetlabs/server/data/<%= EZBake::Config[:real_name] %>/resta
...
-PIDFILE="/var/run/puppetlabs/${realname}/${realname}.pid"
+rundir="/var/run/puppetlabs/${realname}"
+PIDFILE="${rundir}/${realname}.pid"
 
 if [ ! -e "${INSTALL_DIR}/ezbake-functions.sh" ]; then
     echo "Unable to find ${INSTALL_DIR}/ezbake-functions.sh script, failing start." 1>&2
     exit 1
 fi
 
+/usr/bin/install --directory --owner=$USER --group=$GROUP --mode=755 "$rundir"
```
melissa: It doesn't look like we have `$USER` or `$GROUP` set yet

resources/puppetlabs/lein-ezbake/template/pe/ext/redhat/init.suse.erb
```diff
@@ -72,6 +72,7 @@ start() {
 
     export HOME="$(getent passwd ${USER} | cut -d':' -f6)"
 
+    /usr/bin/install --directory --owner=$USER --group=$GROUP --mode=755 "/var/run/puppetlabs/${realname}"
```
melissa: I'm not sure `$USER` or `$GROUP` have been initialized yet in this script

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
camlow325: We inherit them from the sourcing the /etc/sysconfig|default in the upstream app command.  Maybe it would be better to fall back to the default from the Ezbake config in the event they weren't there, though.

resources/puppetlabs/lein-ezbake/template/pe/ext/redhat/init.suse.erb
camlow325: Again, probably inherited from the /etc/sysconfig|default script here.  Looks like there were other references to $USER at least before this change.

melissa: Ahh, right, good point. Thanks for clarifying that

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
melissa: it's up to you, as long as they get set

shrug:
👍

melissa:
Let us know how testing goes, and we'll get this in and released asap

camlow325:
@melissa, @shrug - k, I think this is good to go now.

I did some smoke testing of this with OSS puppetdb from these packages on CentOS 6/7 and Ubuntu 14.04/16.04.  The puppetdb service installed and started properly.  With CentOS 7 and 16.04, I removed the `/var/run/puppetlabs/puppetdb` directory and restarted the VM.  After reboot, the `/var/run/puppetlabs/puppetdb` directory was recreated with puppetdb:puppetdb ownership and the service started properly.

I also smoke tested this with PE puppetdb from these package on SLES 11 and Ubuntu 16.04.  The service started up fine and the tmpfiles.d file for Ubuntu 16.04 had the proper directory name in it.  I removed the `/var/run/puppetlabs/puppetdb` directory on Ubuntu 16.04 and restarted the VM.  On restart, the run directory was recreated with the expected permissions and the service again started properly.

PR #378, opened 2016-10-19
(EZ-104) Restore missing oom command line argument

camlow325:
This commit restores the -XX:OnOutOfMemoryError / kill -9 command line
argument to the Java command line in the start script.  This was
previously unintentionally lost in the conversion of the Java command
line construction from the init scripts and systemd service definitions
over to the start script.

### DIFF
--- a/resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
+++ b/resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
@@ -48,6 +48,7 @@ rm -f "$PIDFILE"
 init_restart_file "$restartfile" || exit $?
 
 ${JAVA_BIN} ${JAVA_ARGS} -Djava.security.egd=/dev/urandom \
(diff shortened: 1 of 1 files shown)


camlow325:
@cprice404, @shrug, @melissa - ping for review, please.  Just a one-liner so hopefully an easy one.

I built this into puppetserver packages at http://builds.delivery.puppetlabs.net/puppetserver/2.6.1.master.SNAPSHOT.2016.10.18T1916.  On CentOS 7 and Ubuntu 14.04, I lowered the max heap size to -Xmx10m and started the puppetserver service.  The service, as expected, failed to start and the process was killed.  The journal / daemon log had messages like the following in each case:

```
# java.lang.OutOfMemoryError: GC overhead limit exceeded
# -XX:OnOutOfMemoryError="kill -9 %p"
#   Executing /bin/sh -c "kill -9 11646"...
```

PR #382, opened 2016-11-04
(EZ-102) Update EZBake to install SysV + SystemD

melissa:
This commit updates EZBake to start installing sysv files on newer
systemd based debian-ish systems. The idea being that users should be
able to switch between the older and newer service management tools.
This is a behavior that has already been established in other packages
that are targeted at newer releases.

This commit also updates EZBake to provide native SystemD files for
Debian 8 Jessie. This had not previously been done because of concerns
over saving state on upgrade from SysV to SystemD. This is no longer a
concern.

### DIFF
--- a/resources/puppetlabs/lein-ezbake/template/pe/ext/debian/postinst.erb
+++ b/resources/puppetlabs/lein-ezbake/template/pe/ext/debian/postinst.erb
@@ -4,7 +4,8 @@
 
 # On upgrade, we should restart the service if it's running
 if [ $1 = 'configure' -a -n $2 ] ; then
-  if [ -d /run/systemd/system ]; then
+  if [ -d '/run/systemd/system' ] ; then
+    # Using systemd
     systemctl daemon-reload >/dev/null 2>&1 || :
     systemctl try-restart <%= EZBake::Config[:project] %>.service ||:
   else
--- a/resources/puppetlabs/lein-ezbake/template/pe/ext/debian/postinst.erb
+++ b/resources/puppetlabs/lein-ezbake/template/pe/ext/debian/postinst.erb
@@ -4,12 +4,7 @@
 
 # On upgrade, we should restart the service if it's running
 if [ $1 = 'configure' -a -n $2 ] ; then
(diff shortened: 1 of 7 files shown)


melissa:
cc/ @haus @underscorgan @mckern Thoughts? I'm still running through test scenarios, so this isn't ready for merge.

melissa:
If you have time to help test, packages are at http://builds.puppetlabs.lan/puppetserver/2.6.1.master.SNAPSHOT.2016.11.04T2126. This link is internal only, so is only accessible via the internal puppet network.

resources/puppetlabs/lein-ezbake/template/pe/ext/debian/postinst.erb
```diff
@@ -4,12 +4,8 @@
 
 # On upgrade, we should restart the service if it's running
 if [ $1 = 'configure' -a -n $2 ] ; then
-  if [ -e "/lib/systemd/system/<%= EZBake::Config[:project] -%>.service" ] ; then
+  f [ -d '/run/systemd/system' ] ; then
```
haus: f => if

haus:
@melissa i'll try to think about this

melissa:
I tested two upgrade scenarios on Debian Jessie.
1) upgrade from the old package with the service running. The service was running post-upgrade as expected.
2) upgrade from the old packages with the service stopped. The service was not running post-upgrade as expected.

I tested a fresh install on Debian Jessie, where I successfully modified the service with puppet.

I tested a fresh install on Ubuntu Xenial 16.04 where I had removed systemd from the system. The service was correctly managed with upstart, and I was able to successfully modify the service with puppet.

Are there any further test cases folks would like to see?

underscorgan:
@melissa I think this sounds reasonable and those look like good test cases.

👍

melissa:
I've updated this PR to pull in @zeha 's commit. Beyond that change, this PR is the same as when I tested it last

mckern:
No real feedback for the work itself. Can removing `jessie` be split into its own commit for the sake of reverting if needed?

melissa:
@mckern updated

underscorgan:
🎉 💯 🎉

PR #384, opened 2016-11-28
(SERVER-1670) Stop autogenerating dependencies for RPM packages

shrug:
Currently the generated RPM packages will automatically generate
a set of requirements based on the contents of the package. This
has become an issue with puppet-server 2.7.0, as adding gems
to the package inadvertently added a dependency on system ruby.

This adds 'AutoReq: 0' to the RPM specs to stop generating automatic
dependency information, and adds 'bash' as the only dependency
that was being automatically added that we still want.

This also adds 'AutoProv: 0' to stop generating a package 'provides'.
This should be unnecessary, but harmless and may save us some pain in
the future if we wind up shipping any shared libs in a package somewhere.

### DIFF
--- a/resources/puppetlabs/lein-ezbake/template/foss/ext/redhat/ezbake.spec.erb
+++ b/resources/puppetlabs/lein-ezbake/template/foss/ext/redhat/ezbake.spec.erb
@@ -81,6 +81,11 @@ Obsoletes:        <%=package-%> <= <%=version-%>-1
 Conflicts:        <%=package-%> <= <%=version-%>-1
 <% end %>
 
+# Don't run the automatic dependency generation, we should be explicitly
+# requiring what we need.
+AutoReq:          0
+AutoProv:         0
+
 BuildRequires:    ruby
 BuildRequires:    /usr/sbin/useradd
 %if %{_with_systemd}
@@ -101,6 +106,7 @@ Requires(postun): systemd
 %endif
 
 Requires:         %{open_jdk}
+Requires:         bash
 # net-tools is required for netstat usage in service unit file
 # See: https://tickets.puppetlabs.com/browse/SERVER-338
 Requires:         net-tools
--- a/resources/puppetlabs/lein-ezbake/template/pe/ext/redhat/ezbake.spec.erb
+++ b/resources/puppetlabs/lein-ezbake/template/pe/ext/redhat/ezbake.spec.erb
@@ -76,6 +76,11 @@ Obsoletes:        <%=package-%> <= <%=version-%>-1
 Conflicts:        <%=package-%> <= <%=version-%>-1
 <% end %>
 
+# Don't run the automatic dependency generation, we should be explicitly
+# requiring what we need.
+AutoReq:          0
+AutoProv:         0
+
 BuildRequires:    puppet-agent
 BuildRequires:    /usr/sbin/useradd
 %if %{_with_systemd}
@@ -100,6 +105,7 @@ Requires(postun): systemd
 
 Requires:         pe-java
 Requires:         pe-puppet-enterprise-release
+Requires:         bash
 # net-tools is required for netstat usage in service unit file
 # See: https://tickets.puppetlabs.com/browse/SERVER-338
 Requires:         net-tools


resources/puppetlabs/lein-ezbake/template/foss/ext/redhat/ezbake.spec.erb
```diff
@@ -81,6 +81,11 @@ Obsoletes:        <%=package-%> <= <%=version-%>-1
 Conflicts:        <%=package-%> <= <%=version-%>-1
 <% end %>
 
+# Don't run the automatic dependency generation, we shoukd be explicitly
```
haus: shoukd => should in both files :)

shrug: Typing is hard.

haus:
looks great. just a typo to fixup

camlow325:
Using an installation of ezbake with these changes, I did a Puppet Server EL-7 ezbake build, http://builds.puppetlabs.lan/puppetserver/2.7.2.master.SNAPSHOT.2016.11.29T0829.  I installed the package from that and ran `rpm -q --requires puppetserver` to compare the dependencies of that package vs. previously built SNAPSHOT from the same branch, http://builds.puppetlabs.lan/puppetserver/2.7.2.master.SNAPSHOT.2016.11.29T1059.

Without AutoReq / AutoProv changes:

```
/bin/bash
/bin/sh
/bin/sh
/bin/sh
/bin/sh
/usr/bin/env
/usr/bin/ruby
config(puppetserver) = 2.7.2.master-0.1SNAPSHOT.2016.11.29T1059.el7
java-1.8.0-openjdk-headless
net-tools
procps
puppet-agent >= 1.6.0
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(FileDigests) <= 4.6.0-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
systemd
systemd
systemd
rpmlib(PayloadIsXz) <= 5.2-1
```

With AutoReq / AutoProv changes:

```
systemd
systemd
systemd
java-1.8.0-openjdk-headless
bash
net-tools
procps
puppet-agent >= 1.6.0
/bin/sh
/bin/sh
/bin/sh
/bin/sh
rpmlib(FileDigests) <= 4.6.0-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(PayloadIsXz) <= 5.2-1
```

The main differences I see are:

1) Several exes that were in the list with AutoReq enabled were gone, including `/usr/bin/ruby`, which is the main one we were hoping to eliminate - which seems good.  In the case of bash, `/bin/bash` is replaced with `bash`, which is what we were hoping for.

/bin/bash
/usr/bin/env
/usr/bin/ruby

2) With AutoReq enabled or not, `/bin/sh` appears in the requires list.  Doesn't seem bad although it did seem a bit surprising.

3) With AutoReq disabled, the following require disappears:

config(puppetserver) = 2.7.2.master-0.1SNAPSHOT.2016.11.29T1059.el7

I don't know what this require is so I'm not sure if we should care if it goes away or not.  @haus, @shrug - WDYT?

melissa:
From my brief research, we shouldn't be worried that the config(puppetserver) dependency disappeared. It seams that was a feature that was meant to allow clean replacement of another package's configuration, but that feature was never fully implemented. However, that answer was from 2011, so things may have changed. That being said, I still don't think it's going to impact our usage

http://lists.rpm.org/pipermail/rpm-maint/2011-June/003058.html is where that answer came from

shrug:
I'm reasonably sure that the config(thing) requires generated by the AutoReq scripts are vestigial in RPM and not actually used  for anything.

I'm basing this largely on the conversation from http://lists.rpm.org/pipermail/rpm-maint/2011-June/003058.html and the fact that it's not really documented in the RPM docs.

shrug:
What @melissa said ^^ :)

camlow325:
Sounds good to me.  Thanks for checking into this!

PR #386, opened 2017-01-12
(CPR-401) Require 'which'

underscorgan:
We add the dependency specifically on '/usr/bin/which' to be flexible
with minimal installers which may provide functionality from different
packages.


shrug:
👍

PR #390, opened 2017-01-13
(EZ-105) Abort when subprocesses exit w/ non-zero code.

waynr:


waynr:
/cc @senior @wkalt this should ensure your CI pipeline fails during the packing job when appropriate rather than continuing into the beaker test job. To get there the following needs to happen:
* [ ] This PR gets merged
* [ ] ezbake `stable` is merged up into `master`
* [ ] ezbake `1.1.7` is released
* [ ] https://github.com/puppetlabs/puppetdb/blob/master/project.clj#L170 and https://github.com/puppetlabs/puppetdb/blob/stable/project.clj#L146 are bumped to `1.1.7`

PR #391, opened 2017-01-24
(maint) Remove cprice404 as maintainer

underscorgan:

PR #392, opened 2017-01-24
remove cprice404

call:


stahnma:
This is an open source project, if @cprice404 wants to stay on, seems like he can.

underscorgan:
see also, #391

call:
My understanding was that only Puppet employees belong in MAINTAINERS files. cc @theshanx @rick @stahnma

underscorgan:
I'm going to close this since it's a dupe of #391 and there are duplicate conversations now.

PR #395, opened 2017-03-10
(SERVER-1742) Improve init script behavior for service stop timeouts

shrug:
Currently, the SysV init script 'restart' command will try to start the service even if the stop did not successfully complete.

This updates the service stop command to use the stop timeout value in the 'kill' signal in addition to the 'term' signal, and updates the init scripts to only attempt to start the service if the stop completes successfully in a 'restart'


melissa:
Were you going to add a note to users about extending the timeout if their heap size is too big?

PR #397, opened 2017-03-13
(SERVER-1742) Add a comment to defaults file about stop timeouts

shrug:
...and heap sizes


shrug:
ping @camlow325 Are you ok with the wording on this? I'm not in love with it, but I didn't want to write an essay in the comments either.

camlow325:
@shrug This looks fine to me.  I'm also fine with not mentioning anything about GC / large heap here and seeing if the 60 second default is sufficient for a `kill -9` without having to bother users with the extra detail.  I don't have any problem with this, though, if you want to go with it.

camlow325:
Woops, accidentally closed the PR instead of just commenting on it.  🤕 Just reopened it.

shrug:
Fair enough, it's also probably a fairly rare case. I'll close this and we can add it later (and to docs) if it becomes necessary

PR #396, opened 2017-03-10
(RE-8191) Remove java7 references from package builds

shrug:

PR #399, opened 2017-03-20
(SERVER-1763) Add ca-certificates as build dependency

haus:
Now that puppetserver has started bundling gems into the ezbake build,
java's lack of truststore anchors in our build has become a problem. In
jruby, simply requiring openssl will fail if there are no trust anchors.
This is the case on sles12. To avoid this problem, this commit adds the
correct ca-certificates package for el and for sles 12.

### DIFF
--- a/resources/puppetlabs/lein-ezbake/template/foss/ext/redhat/ezbake.spec.erb
+++ b/resources/puppetlabs/lein-ezbake/template/foss/ext/redhat/ezbake.spec.erb
@@ -27,6 +27,7 @@
 %global _systemd_sles  0
 %global _old_el        0
 %global _sles          0
+%global _certs_package  ca-certificates
 
 %if 0%{?fedora} >= 18 || 0%{?rhel} >= 7
 %global _with_systemd 1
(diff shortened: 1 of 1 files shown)


camlow325:
@melissa, @underscorgan, or @shrug - ok with you if we merge this and release an ezbake version with this change in it.  This is currently causing our puppetserver CI to fail for SLES12 so we'd like to be able to fix this soon.  Thanks!

camlow325:
Although I guess one question would be what branch we might want this to release from.  If we're not quite ready to start doing releases from the master branch yet, e.g., because we want the changes in #396 to land first, then maybe it would be better to retarget this as stable and do a Y release from there.  @melissa, @underscorgan, or @shrug - WDYT?

melissa:
👍 

As to your question on where to merge this, I'm not sure what the state of the the java7 removal effort, or what the plan is for ezbake versioning there. @shrug might have more concrete details?

shrug:
👍 

As far as a target, we'll probably want this in both stable and master. We can still merge master -> stable until #396 lands. Or retarget at Stable.

rlinehan:
I made a jira ticket: SERVER-1763. Don't know if anyone cares, but it's there.

camlow325:
Since we have multiple 👍s I'm going to merge this.  Looks like it's already been retargeted at stable, which seems good.

PR #401, opened 2017-04-03
(maint) Towards Transparent, Reproducible Snapshot Promotions

aperiodic:
Change EZBake's stage command to:
1) Deploy an artifact to the configured snapshots repository when staging a project with a snapshot version.
2) List the deployed snapshot artifact's version as the project's version number in the ezbake.manifest & project_data.yaml files.
3) Resolve all dependencies with 'SNAPSHOT' versions to get a deployed snapshot artifact from the repository and list that artifact's version number in the ezbake.manifest & project_data.yaml. If no deployed snapshot artifacts can be found for the listed snapshot version, then an error is thrown and staging is aborted, to prevent unreproducible builds.
4) List each dependency's group as well as its name in the ezbake.manifest & project_data.yaml

See the individual commit mesages for more details.

### DIFF
--- a/src/puppetlabs/ezbake/core.clj
+++ b/src/puppetlabs/ezbake/core.clj
@@ -523,10 +523,14 @@ Dependency tree:
   (fn [action & args] action))
 
 (defmethod action "stage"
+  ;; note that the `lein-project` arg gets shadowed a few lines down to add full
+  ;; snapshot versions of dependencies
   [_ lein-project build-target]
   (let [deployed-version (if (deputils/snapshot-version? (:version lein-project))
                            (deploy-snapshot lein-project)
-                           (:version lein-project))]
+                           (:version lein-project))
+        lein-project (update lein-project :dependencies (partial deputils/expand-snapshot-versions
+                                                                 lein-project))]
     (let [template-dir (get-template-file build-target)
           uberjar-name (:uberjar-name lein-project)]
       (uberjar/uberjar lein-project)
--- a/src/puppetlabs/ezbake/core.clj
+++ b/src/puppetlabs/ezbake/core.clj
@@ -7,6 +7,7 @@
             [clj-time.local :as local-time]
             [stencil.core :as stencil]
             [leiningen.core.main :as lein-main]
+            [leiningen.deploy :as deploy]
             [leiningen.uberjar :as uberjar]
             [schema.core :as schema]
             [schema.utils :as schema-utils]
@@ -286,6 +287,17 @@ Dependency tree:
         (deputils/cp-files-of-type lein-project "doc"
                                    docs-prefix get-out-dir-for-doc-file)))
 
+(defn deploy-snapshot
+  "Given a project map with a snapshot version, deploy to the configured
+  snapshots repository and return the version of the deployed artifact. If given
+  a map with a non-snapshot version, does nothing and returns nil."
+  [lein-project]
+  (when (deputils/snapshot-version? (:version lein-project))
+    (-> (deploy/deploy lein-project)
+      .getArtifacts
+      (nth 0)
+      .getVersion)))
+
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Upstream EZBake config handling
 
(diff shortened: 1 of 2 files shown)


shrug:
I'm not sure I understand this well enough to review. Paging Dr. @haus ?

If I'm doing local test builds, they'll now be deployed to the repository? Is that desirable? How does that impact people building outside of our network?

src/puppetlabs/ezbake/core.clj
```diff
@@ -286,6 +287,17 @@ Dependency tree:
...
 
+(defn deploy-snapshot
+  "Given a project map with a snapshot version, deploy to the configured
+  snapshots repository and return the version of the deployed artifact. If given
+  a map with a non-snapshot version, does nothing and returns nil."
+  [lein-project]
+  (when (deputils/snapshot-version? (:version lein-project))
+    (-> (deploy/deploy lein-project)
+      .getArtifacts
+      (nth 0)
```
haus: aether's javadocs seem to be inaccessible right now. does deploy return just the artifacts that were deployed?

src/puppetlabs/ezbake/dependency_utils.clj
```diff
@@ -214,3 +214,7 @@
                     out-dir-fn
                     lein-project)
            deps))))
+
+(defn snapshot-version?
+  [version]
+  (boolean (re-find #"SNAPSHOT" version)))
```
haus: pomegranate uses a slightly more strict regex (includes a `-`), so maybe we should add a `-`? (https://github.com/cemerick/pomegranate/blob/e44f45fa3abc29dbc0ab88f98b9bbbada7819128/src/main/clojure/cemerick/pomegranate/aether.clj#L831). Probably not a big deal either way.

aperiodic: seems good to agree on that. I'll change it

src/puppetlabs/ezbake/dependency_utils.clj
```diff
@@ -30,6 +30,47 @@
...
+                           :coordinates [coords]
+                           :repositories (:repositories lein-project)
+                           :mirrors (:mirrors lein-project)
+                           :local-repo (:local-repo lein-project))
+                       first
+                       meta
+                       :result
+                       .getArtifact
+                       .getVersion)]
+    (if-not (re-find #"SNAPSHOT$" full-version)
```
haus: same as other comment about SNAPSHOT regex applies here

src/puppetlabs/ezbake/dependency_utils.clj
```diff
@@ -30,6 +30,47 @@
...
+      full-version
+      (throw
+        (Exception.
```
haus: total nitpick, but it would be good to throw something slightly more specific here, like IllegalArgumentException or IllegalStateException maybe?

haus:
This looks pretty reasonable to me. Would you mind gisting or pasting a before and after manifest and package description?

src/puppetlabs/ezbake/core.clj
aperiodic: Yes. `.deploy` returns a [DeployResult](https://sonatype.github.io/sonatype-aether/apidocs/index.html?org/sonatype/aether/util/version/GenericVersionScheme.html) which only contains the artifacts that were deployed.

haus: okay, i was thinking about the case where there is a test jar deployed along with the main jar. in that case i think the versions would still be the same, so it wouldn't really matter which artifact we got.

src/puppetlabs/ezbake/dependency_utils.clj
aperiodic: I looked through the standard Exception subclasses and couldn't see anything that looked appropriate. It's not an IllegalArgumentException since the problem isn't the argument, it's that no versions matching the snapshot have been deployed. I suppose I could change it to `IllegalStateException`, which is accurate enough if you squint and handwave a bit, in order to make it easier to catch just this problem.

aperiodic:
@haus I made the requested changes; generating before & after manifests now

aperiodic:
@haus here's a diff between the ezbake.manifest for pe-console-services before & after this PR's changes: https://gist.github.com/aperiodic/b16e3c447a692bdfd4f710caecd26688

aperiodic:
@haus so can we merge this?

haus:
I don't see why not.

PR #402, opened 2017-04-06
(RE-8694) Add `platform_repos` entry to build_defaults

mwaggett:
This commit adds a `platform_repos` entry to build_defaults.yaml, similar to the one in puppet-agent. When built, this info will get put into the \<ref\>.yaml file created by packaging, which can then be used by beaker for testing.

### DIFF
--- a/resources/puppetlabs/lein-ezbake/template/foss/ext/build_defaults.yaml
+++ b/resources/puppetlabs/lein-ezbake/template/foss/ext/build_defaults.yaml
@@ -17,3 +17,20 @@ build_ips: FALSE
 apt_host: 'apt.puppetlabs.com'
 apt_repo_url: 'http://apt.puppetlabs.com'
 apt_repo_path: '/opt/repository/incoming'
+platform_repos:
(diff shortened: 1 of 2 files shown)


underscorgan:
I think I'm leaning towards having the repos be based on platforms expected to be built for PE / FOSS. It doesn't feel like it would be useful for beaker to know how to test FOSS puppetserver for PE only platforms or vice versa?

mwaggett:
@underscorgan yeah, that's what I was thinking. My only hesitation is that, since these are in ezbake, rather than a specific project, if different projects are building for different platforms, there could be some confusion? but I don't know if that's a likely scenario.

underscorgan:
AFAIK ezbake doesn't really support building different platforms based on project unless they are using a different version of EZBAKE. The projects will build based on the cows and final_mocks parameters. Those can be overwritten by environment variable but I don't know that any build pipelines actually do that?

mwaggett:
After feedback from Sean, we have decided that this isn't the best direction for this work.

PR #403, opened 2017-04-11
(EZ-108) Don't deploy if `EZBAKE_NODEPLOY` set

aperiodic:
Change the `stage` action to avoid deploying to the snapshots repository
if the `EZBAKE_NODEPLOY` environment variable is set to any value
(including `false` or `0` or `null`), and mention this in the README

PR #405, opened 2017-04-12
(maint) Backport Snapshot-Version Changes & Full Dependency Names

aperiodic:
Backport the recently merged changes on master to list specific snapshot versions included in a package and to have the full dependency symbol (including the group) in the description.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,13 +156,22 @@ lein with-profile ezbake ezbake stage
 ```
 
 This will create an ephemeral git repository at `./target/staging` with staged
-templates ready for consumption by the build step. If the project being staged
-has a SNAPSHOT version, then a snapshot build will be deployed to the project's
-configured snapshots repository (typically our internal nexus repository
-server at `nexus.delivery.puppetlabs.net`), in order to ensure our builds are
-reproducible. This can be avoided by setting the `EZBAKE_NODEPLOY` environment
-variable to any value. (If you set this environment variable in a build
-pipeline, you will be severely punished by the Angel of Build Reproducibility.)
+templates ready for consumption by the build step.
+
+If the project being staged has a SNAPSHOT version, then a snapshot build will
+be deployed to the project's configured snapshots repository (typically our
+internal nexus repository server at `nexus.delivery.puppetlabs.net`), in order
+to ensure our builds are reproducible. This can be avoided by setting the
+`EZBAKE_NODEPLOY` environment variable to any value.
+
+If the project or any of its dependencies have a SNAPSHOT version, but
+a deployed snapshot artifact matching that version cannot be found in the
+configured repositories (for example, because they have only been installed
+locally), then EZBake will throw an error to prevent an unreproducible build. If
+you're testing local changes, you can avoid this check by setting the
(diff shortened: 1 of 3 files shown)


underscorgan:
Can you please also include https://github.com/puppetlabs/ezbake/pull/403 ?

aperiodic:
@underscorgan oh, of course. Updated.

aperiodic:
@underscorgan is this good to go now? Does it need more changes?

shrug:
With EZBAKE_NODEPLOY set, the stage command doesn't deploy the artifact, howver the build command fails if there's not a deployed artifact - is that intentional? it makes testing changes locally more difficult.

aperiodic:
@shrug yes. I think the extra step it adds to testing local changes is worth being sure that we're not making unreproducible builds. There are a few different ways to make it [easier to deploy the snapshot](https://github.com/technomancy/leiningen/blob/master/doc/DEPLOY.md#authentication), including just deploying once upfront and then setting EZBAKE_NODEPLOY

shrug:
So for example, at the moment I'm trying to run test puppet-server builds in order to test some ezbake changes. Do we expect those to be deployed? I'll need to configure credentials to be able to deploy artifacts if that's the case

aperiodic:
@shrug if it's building snapshot versions, then yes. Like most of our clojure projects, puppetserver is [configured to source deployment credentials from the environment](https://github.com/puppetlabs/puppetserver/blob/master/project.clj#L6), and the correct environment variables are already set [in the packaging-clj-lein-ezbake Jenkins job](https://github.com/puppetlabs/ci-job-configs/blob/master/resources/scripts/packaging-clj-lein-ezbake.sh).

camlow325:
Like @shrug, I would really like to have a way to disable the `get-full-snapshot-version` functionality occasionally for local testing - maybe through the use of an extra `EZBAKE_NO_RESOLVE_SNAPSHOTS` variable or something if we don't want to overload the use of the `EZBAKE_NODEPLOY` variable.  I frequently find myself running "ezbake stage" locally for one-off, iterative puppetserver work and would like to just validate that content is staged properly by ezbake without having to rely upon having deployed something to an external repository.  I can hack around this for now by setting up an internal local file repository and doing a local deploy of puppetserver to that so that the stage command is able to resolve the dependency.  That process has been pretty tedious so far for me, though, and so just having a "buyer beware" way to turn it off at the command line would be very helpful.

aperiodic:
@shrug @camlow325 I've added a commit that lets you set the `EZBAKE_ALLOW_UNREPRODUCIBLE_BUILDS` environment variable to bypass the deployed snapshot check.

camlow325:
@aperiodic Thanks for doing that.  I think that is a good compromise.

shrug:
@aperiodic thanks, mind adding that to the README and we'll get this in?

aperiodic:
@shrug done

shrug:
@camlow325 Did you want to take another look, or are you ok to merge?

camlow325:
@shrug Feel free to merge if y'all are ok with it.  I just wanted to weigh in on the one part about resolving snapshot versions, which I'm good with.

PR #410, opened 2017-04-24
(SERVER-1772) Install additional uberjars

jpinsonault:
This commit adds support for installing additional uberjars retrieved from
maven coordinates in a project.clj file

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ Note that the symble `:ezbake` is not strictly necessary here.
 
 #### Composite
 
-Composite ezbake projects usually do not define their own services but rather
+Composite EZBake projects usually do not define their own services but rather
 provide a list of dependencies which themselves define TK services.  Because of
 this it is not strictly necessary to define a profile such as `:ezbake` shown
 above; although it is conceivable that such a composite project may define its
@@ -142,7 +142,7 @@ in architectural diagrams. Just look at the Leaning Tower of Pisa.
 
 #### Additional Uberjars
 
-Some projects might have a use case for automatically fetching an versioned
+Some projects might have a use case for automatically fetching a versioned
 jar from an external repository that needs to be placed into the package and be
 available after installation.
 
@@ -161,12 +161,12 @@ same directory as your project's own uberjar after installation.
 The filename of the built uberjar is determined by what value `:uberjar-name`
 is set to in its `project.clj`
 
-ezbake will attempt to resolve the coordinates of these external uberjars using
+EZBake will attempt to resolve the coordinates of these external uberjars using
 the repositories specified in the `:repositories` key of your project.clj
 
 #### cli-defaults.sh
 `cli-defaults.sh` is intended to be a file from which default shell script
-variables can be defined by a project, and be loaded by ezbake's `cli-app`
+variables can be defined by a project, and be loaded by EZBake's `cli-app`
 script. That is, the script which is run when commands like
 `service my-project start` are called.
 
--- a/README.md
+++ b/README.md
@@ -140,6 +140,44 @@ above; although it is conceivable that such a composite project may define its
 own services, it is unlikely and ill-advised because no one likes blurred lines
 in architectural diagrams. Just look at the Leaning Tower of Pisa.
 
+#### Additional Uberjars
+
+Some projects might have a use case for automatically fetching an versioned
+jar from an external repository that needs to be placed into the package and be
+available after installation.
+
+The `:additional-uberjars` setting can be set to a list of project coordinates
+that will be resolved and have uberjars built from them. For example:
+
+```clojure
+{:lein-ezbake {
+  :additional-uberjars [[puppetlabs/puppetserver "2.7.2"]]
+...}}
+```
+
+This would result in a puppetserver uberjar being built, and the jar being placed in the
+same directory as your project's own uberjar after installation.
+
+The filename of the built uberjar is determined by what value `:uberjar-name`
+is set to in its `project.clj`
+
+ezbake will attempt to resolve the coordinates of these external uberjars using
+the repositories specified in the `:repositories` key of your project.clj
+
+#### cli-defaults.sh
+`cli-defaults.sh` is intended to be a file from which default shell script
+variables can be defined by a project, and be loaded by ezbake's `cli-app`
+script. That is, the script which is run when commands like
+`service my-project start` are called.
+
+This differs from the `default` files, which also get loaded by `cli-app`, in
+that a project has no control over what goes into `default`. `cli-defaults.sh`
+is under the control of your project.
+
+For `cli-defaults.sh` to be used, it simply needs to exist at
+`resources/ext/cli_defaults/cli-defaults.sh.erb` in your project. Since it is
+an erb template, it will have access to the variables in `ezbake.rb`
+
 ### Running
 
 Running ezbake works much like any other Leiningen plugin or built-in task.
--- a/resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
+++ b/resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
@@ -10,7 +10,7 @@ realname="<%= EZBake::Config[:real_name] %>"
 rundir="/var/run/puppetlabs/${realname}"
 PIDFILE="${rundir}/${realname}.pid"
 
-cli_defaults=${INSTALL_DIR}/cli/apps/cli-defaults.sh
+cli_defaults=${INSTALL_DIR}/cli/cli-defaults.sh
 
 if [ ! -e "${INSTALL_DIR}/ezbake-functions.sh" ]; then
     echo "Unable to find ${INSTALL_DIR}/ezbake-functions.sh script, failing start." 1>&2
--- a/resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
+++ b/resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
@@ -10,6 +10,8 @@ realname="<%= EZBake::Config[:real_name] %>"
 rundir="/var/run/puppetlabs/${realname}"
 PIDFILE="${rundir}/${realname}.pid"
 
+cli_defaults=${INSTALL_DIR}/cli/apps/cli-defaults.sh
+
 if [ ! -e "${INSTALL_DIR}/ezbake-functions.sh" ]; then
     echo "Unable to find ${INSTALL_DIR}/ezbake-functions.sh script, failing start." 1>&2
     exit 1
@@ -47,10 +49,10 @@ rm -f "$PIDFILE"
 
 init_restart_file "$restartfile" || exit $?
 
-CLASSPATH=${INSTALL_DIR}/puppet-server-release.jar
+CLASSPATH=${INSTALL_DIR}/<%= EZBake::Config[:uberjar_name] %>
 
-if [ -e "${INSTALL_DIR}/cli/apps/cli-defaults.sh" ]; then
-  . ${INSTALL_DIR}/cli/apps/cli-defaults.sh
+if [ -e "$cli_defaults" ]; then
+  . $cli_defaults
   if [ $? -ne 0 ]; then
     echo "Unable to initialize cli defaults, failing start." 1>&2
     exit 1
@@ -59,7 +61,7 @@ fi
 
 ${JAVA_BIN} ${JAVA_ARGS} -Djava.security.egd=/dev/urandom \
   -XX:OnOutOfMemoryError="kill -9 %p" \
-  -cp ${CLASSPATH} \
+  -cp "${CLASSPATH}" \
   clojure.main \
   -m <%= EZBake::Config[:main_namespace] %> \
   --config "${CONFIG}" \
--- a/resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
+++ b/resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
@@ -47,9 +47,19 @@ rm -f "$PIDFILE"
 
 init_restart_file "$restartfile" || exit $?
 
+CLASSPATH=${INSTALL_DIR}/puppet-server-release.jar
+
+if [ -e "${INSTALL_DIR}/cli/apps/cli-defaults.sh" ]; then
+  . ${INSTALL_DIR}/cli/apps/cli-defaults.sh
+  if [ $? -ne 0 ]; then
+    echo "Unable to initialize cli defaults, failing start." 1>&2
+    exit 1
+  fi
+fi
+
 ${JAVA_BIN} ${JAVA_ARGS} -Djava.security.egd=/dev/urandom \
   -XX:OnOutOfMemoryError="kill -9 %p" \
-  -cp ${INSTALL_DIR}/<%= EZBake::Config[:uberjar_name] %> \
+  -cp ${CLASSPATH} \
   clojure.main \
   -m <%= EZBake::Config[:main_namespace] %> \
   --config "${CONFIG}" \
--- a/resources/puppetlabs/lein-ezbake/template/global/install.sh.erb
+++ b/resources/puppetlabs/lein-ezbake/template/global/install.sh.erb
@@ -28,6 +28,7 @@ uxbindir=${uxbindir:=/opt/puppetlabs/bin}
 # symlinks of server binaries
 symbindir=${symbindir:=/opt/puppetlabs/server/bin}
 app_prefix=${app_prefix:=/opt/puppetlabs/server/apps/${real_name}}
+dest_apps_dir="${DESTDIR}${app_prefix}"
 app_data=${app_data:=/opt/puppetlabs/server/data/${real_name}}
 app_logdir=${app_logdir:=/var/log/puppetlabs/${real_name}}
 system_config_dir=${system_config_dir:=${app_prefix}/config}
@@ -131,16 +132,16 @@ function task_install_source_deb_systemd {
 # Install the ezbake package software. This step is used during RPM &
 # Debian packaging during the 'install' phases.
 function task_install {
-    install -d -m 0755 "${DESTDIR}${app_prefix}"
+    install -d -m 0755 "${dest_apps_dir}"
     install -d -m 0770 "${DESTDIR}${app_data}"
-    install -m 0644 <%= EZBake::Config[:uberjar_name] %> "${DESTDIR}${app_prefix}"
-    install -m 0755 ext/ezbake-functions.sh "${DESTDIR}${app_prefix}"
-    install -m 0644 ext/ezbake.manifest "${DESTDIR}${app_prefix}"
+    install -m 0644 <%= EZBake::Config[:uberjar_name] %> "${dest_apps_dir}"
+    install -m 0755 ext/ezbake-functions.sh "${dest_apps_dir}"
+    install -m 0644 ext/ezbake.manifest "${dest_apps_dir}"
     install -d -m 0755 "${DESTDIR}${projconfdir}/conf.d"
 
 <% if EZBake::Config[:additional_uberjars]  -%>
   <% EZBake::Config[:additional_uberjars].each do |jar| -%>
-    install -m 0644 <%= jar %> "${DESTDIR}${app_prefix}"
+    install -m 0644 <%= jar %> "${dest_apps_dir}"
   <% end -%>
 <% end -%>
 
@@ -157,12 +158,12 @@ function task_install {
     install -m 0644 <%= config_file %> "${DESTDIR}${projconfdir}/<%= config_file.sub(/^ext\/config\//, "") %>"
 <% end -%>
 
-    install -d -m 0755 "${DESTDIR}${app_prefix}/scripts"
-    install -m 0755 install.sh "${DESTDIR}${app_prefix}/scripts"
+    install -d -m 0755 "${dest_apps_dir}/scripts"
+    install -m 0755 install.sh "${dest_apps_dir}/scripts"
 
 <% if ! EZBake::Config[:cli_app_files].empty? -%>
-    install -d -m 0755 "${DESTDIR}${app_prefix}/cli"
-    install -d -m 0755 "${DESTDIR}${app_prefix}/cli/apps"
+    install -d -m 0755 "${dest_apps_dir}/cli"
+    install -d -m 0755 "${dest_apps_dir}/cli/apps"
     install -d -m 0755 "${DESTDIR}${bindir}"
     install -m 0755 "ext/bin/${real_name}" "${DESTDIR}${bindir}/${real_name}"
     install -d -m 0755 "${DESTDIR}${symbindir}"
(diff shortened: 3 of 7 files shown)


src/puppetlabs/ezbake/core.clj
```diff
@@ -507,26 +510,28 @@ Dependency tree:
...
-       :uberjar-name  (:uberjar-name lein-project)
-       :is-pe-build   (format "%s" (= (get-local-ezbake-var lein-project :build-type "foss") "pe"))
-       :repo-name     (format "%s" (get-local-ezbake-var lein-project :repo-target ""))})))
+      {:project (:name lein-project)
+       :summary (:description lein-project)
+       :description (format "%s (%s)"
+                            (:description lein-project)
+                            (deputils/generate-manifest-string lein-project))
+       :uberjar-name (:uberjar-name lein-project)
+       :additional-uberjars (mapv (fn [filename] {:uberjar filename}) additional-uberjars)
```
jpinsonault: This is the magic bit to get mustache to work

jpinsonault:
Notes on how I tested it:

This is my cli-defaults.sh.erb for puppetserver, in the `resources/ext/cli/` dir

```
INSTALL_DIR="/opt/puppetlabs/server/apps/<%= EZBake::Config[:real_name] %>"

JRUBY_JAR=${JRUBY_JAR:-${INSTALL_DIR}/jruby-1_7.jar}

if [ ! -e "$JRUBY_JAR" ]; then
  echo "Unable to find specified JRUBY_JAR: ${JRUBY_JAR}" 1>&2
  return 1
fi

CLASSPATH="${CLASSPATH}:${JRUBY_JAR}"
```

and then I had to modify the `install-vendored-gems.sh` file to run with the jruby jar

`java -cp puppet-server-release.jar:jruby-1_7.jar clojure.main -m puppetlabs.puppetserver.cli.gem --config jruby.conf -- install ${gem_name} --version ${gem_version}`

and then of course I lein installed your `jruby-deps` branches and added this to my `:lein-ezbake` config

```
:additional-uberjars [[puppetlabs/jruby-deps "9.1.8.0-1-SNAPSHOT"]
                                      [puppetlabs/jruby-deps "1.7.26-1-SNAPSHOT"]]
```

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
```diff
@@ -47,9 +47,19 @@ rm -f "$PIDFILE"
 
 init_restart_file "$restartfile" || exit $?
 
+CLASSPATH=${INSTALL_DIR}/puppet-server-release.jar
```
camlow325: Could this just be:

```
CLASSPATH=${INSTALL_DIR}/<%= EZBake::Config[:uberjar_name] %>
```

jpinsonault: Yeah I thought I had changed it to that already

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
```diff
@@ -47,9 +47,19 @@ rm -f "$PIDFILE"
...
+
+if [ -e "${INSTALL_DIR}/cli/apps/cli-defaults.sh" ]; then
```
camlow325: I'd like to try to find a way to not have this file end up under the ../apps subdirectory so that the cli-app script doesn't try to treat the cli-defaults.sh script like another app, like for purposes of showing the available apps when the '-h' argument is passed in.  Not sure how easy that will be to do though?

jpinsonault: Agreed. Another option I don't like very much is changing the `cli-app.erb` script to be aware of `cli-defaults.sh`

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
```diff
@@ -47,9 +47,19 @@ rm -f "$PIDFILE"
...
+  if [ $? -ne 0 ]; then
+    echo "Unable to initialize cli defaults, failing start." 1>&2
+    exit 1
+  fi
+fi
+
 ${JAVA_BIN} ${JAVA_ARGS} -Djava.security.egd=/dev/urandom \
   -XX:OnOutOfMemoryError="kill -9 %p" \
-  -cp ${INSTALL_DIR}/<%= EZBake::Config[:uberjar_name] %> \
+  -cp ${CLASSPATH} \
```
camlow325: Might be good to put double-quotes around this in case the sourced script were to ever include spaces for some reason.

```
-cp "${CLASSPATH}"
```

resources/puppetlabs/lein-ezbake/template/global/install.sh.erb
```diff
@@ -167,6 +173,11 @@ function task_install {
 <% EZBake::Config[:cli_app_files].each do |cli_app_file| -%>
     install -m 0755 <%= cli_app_file %> "${DESTDIR}${app_prefix}/cli/apps/<%= cli_app_file.sub(/^ext\/cli\//, "") %>"
 <% end -%>
+
+if [ -e "<%= EZBake::Config[:cli_defaults_file] %>" ]; then
+    install -m 0755 <%= EZBake::Config[:cli_defaults_file] %> "${DESTDIR}${app_prefix}/cli/apps/<%= EZBake::Config[:cli_defaults_file] %>"
```
camlow325: Maybe here is where we could have the `cli_defaults_file` be installed to .../cli instead of .../cli/apps - so that the cli-app script doesn't interpret it as an app?

src/puppetlabs/ezbake/core.clj
```diff
@@ -1,19 +1,22 @@
...
+  (:import (java.io File InputStream InputStreamReader)
+           (java.util.jar JarEntry JarFile))
   (:require [me.raynes.fs :as fs]
             [clojure.java.io :as io]
             [clojure.string :as str]
             [clj-time.local :as local-time]
             [stencil.core :as stencil]
             [leiningen.core.main :as lein-main]
             [leiningen.deploy :as deploy]
+            [leiningen.core.project :as lein-project]
```
camlow325: Looks like this may not be used?

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
jpinsonault: good call

src/puppetlabs/ezbake/core.clj
```diff
@@ -499,25 +510,92 @@ Dependency tree:
...
+  [lein-project]
+  (when-let [dependencies-vector (get-in lein-project [:lein-ezbake :additional-uberjars])]
+    dependencies-vector))
+
+(schema/defn resolve-dependency! :- File
+  "Resolves a single dependency and returns a File pointing to it's jar"
+  [[project-symbol version :as dependency-coordinates]]
+  (lein-main/info "Resolving dependency for " dependency-coordinates)
+  (let [project {:dependencies [dependency-coordinates]
+                 :repositories [["releases" {:url "https://clojars.org/repo"}]]}
```
camlow325: We talked about looking into whether we could just use the value from the `:repositories` key in the `defproject` that the ezbake build was performed from.  Hopefully, that would have the complete list of info for repositories that we might expect to be able to retrieve dependencies from.

src/puppetlabs/ezbake/core.clj
```diff
@@ -499,25 +510,92 @@ Dependency tree:
...
+        ; resolved-dependencies will be a list of jars, one for each dependency
+        ; in the project, including the jar for this project
+        resolved-dependencies (lein-classpath/resolve-managed-dependencies :dependencies nil project)
+        jar-regex (re-pattern (format "%s-%s.jar" (name project-symbol) version))
+        jar-file (first (filter #(re-find jar-regex (.getName %))
```
camlow325: Is the first one returned always guaranteed to be the "main" uberjar that we're building from the base project?

src/puppetlabs/ezbake/core.clj
jpinsonault: I think that's a good way to go. I've got those changes locally already

src/puppetlabs/ezbake/core.clj
```diff
@@ -499,25 +510,92 @@ Dependency tree:
...
+  [[project-symbol version :as dependency-coordinates]]
+  (let [dependency-jar (resolve-dependency! dependency-coordinates)
+        destination-dir (format "%s/%s-%s"
+                                additional-uberjar-checkouts-dir
+                                (name project-symbol)
+                                version)
+        project-file (extract-project-from-jar! dependency-jar destination-dir)
+        project-map (-> project-file
+                        project/read
+                        (update :dependencies conj dependency-coordinates))]
```
camlow325: For future maintainer, might be good to add a comment here about why this is necessary.  It does make sense to me, though, based on our prior conversations about it.

src/puppetlabs/ezbake/core.clj
jpinsonault: `filter` should only ever return one match as far as I can figure it. I couldn't think of a way that `<project>-<version>.jar` could match anything other than the one we want. Maybe the regex should be tightened up a bit though

Happy to do something else though. `lein-classpath/resolve-managed-dependencies` is designed to resolve lots of dependencies, not just fetch a single jar. We could write some new code modeled off the guts of `lein-classpath/resolve-managed-dependencies` but I don't know how much effort that would be or what kind of complexity we'd be getting ourselves into

src/puppetlabs/ezbake/core.clj
jpinsonault: Yeah good point. There's probably a few places it needs better comments

src/puppetlabs/ezbake/core.clj
camlow325: oh, I see.  I was misreading that this was just called once for each set of coordinates for an uberjar to be built.  What you have here makes sense to me.

camlow325:
One of the things I'd mentioned in the description for the JIRA ticket, [SERVER-1772](https://tickets.puppetlabs.com/browse/SERVER-1772), that would be good to look at as part of this work is updating the logic which generates the ezbake.manifest file to cover the additional uberjars.  Currently, I believe just the versioning info for the main project uberjar is captured in the ezbake.manifest file.  I'd be fine with adding on this part of the work in a separate PR if you'd prefer.  What do you think?

jpinsonault:
Pushed some commits which

* Integrate the classpath changes into foreground.erb
* Moves cli-defaults.sh from `.../cli/apps/` to `.../cli/` once installed
  * It's now expected to be at `resources/ext/cli_defaults/cli-defaults.sh.erb`
* Get the repositories list from the host project for additional uberjar dependency resolution
* Added some docs (that could probably be improved?)

Might do updating the ezbake.manifest in another PR

src/puppetlabs/ezbake/core.clj
```diff
@@ -32,9 +34,12 @@
 (def config-dir "ext/config/")
 (def system-config-dir "ext/system-config/")
 (def shared-cli-apps-prefix "ext/cli/")
+(def shared-cli-defaults-prefix "ext/cli_defaults/")
 (def docs-prefix "ext/docs/")
 (def build-scripts-prefix "ext/build-scripts/")
 (def terminus-prefix "puppet/")
+(def additional-uberjar-checkouts-dir "target/uberjars")
+(def cli-defaults-filename "ext/cli_defaults/cli-defaults.sh.erb")
```
camlow325: Very small nit - might be good to concatenate the value for `shared-cli-defaults-prefix` into the construction of this variable so that the "ext/cli_defaults/" part only appears once.

jpinsonault:
Pushed a commit to add info to the ezbake.manifest file

camlow325:
With the latest changes, I noticed that, when testing with a Puppet Server build, that the `ezbake.manifest` file includes a couple of dependencies for the jruby-9k.jar and jruby-1_7.jar which I don't believe are or should be there:

```
Additional uberjar dependencies:

jruby-9k.jar:

[clojure-complete "0.2.4" :exclusions [[org.clojure/clojure]]]
[org.clojure/tools.nrepl "0.2.12" :exclusions [[org.clojure/clojure]]]
```

Do you see the same thing?

I don't see these extra dependencies in the listing for the primary jar.  Is there a way we could get rid of them for the additional uberjar info listing?

jpinsonault:
Pushed a commit to make ezbake use the uberjar profile when building the additional uberjars. The extra dependencies are brought in when the default profile is used, so this should fix that

camlow325:
This all looks good to me with the most recent changes.  The ezbake.manifest no longer has the superfluous `clojure-complete` and `org.clojure/tools.nrepl` dependencies for the additional uberjars, which all seems good.

camlow325:
@haus Any chance you'd have time to review this PR soon?  Would be nice to get @shrug, @melissa, and/or @underscorgan to weigh in on this, too.  Where I think you have the most experience among us with the Clojure-related code in ezbake, though, I'd especially appreciate it if you could take a look at this.

src/puppetlabs/ezbake/core.clj
```diff
@@ -519,6 +528,69 @@ Dependency tree:
        :is-pe-build   (format "%s" (= (get-local-ezbake-var lein-project :build-type "foss") "pe"))
        :repo-name     (format "%s" (get-local-ezbake-var lein-project :repo-target ""))})))
 
+(schema/defn get-additional-uberjars
+  [lein-project]
+  (when-let [dependencies-vector (get-in lein-project [:lein-ezbake :additional-uberjars])]
+    dependencies-vector))
+
+(schema/defn resolve-dependency! :- File
+  "Resolves a single dependency and returns a File pointing to it's jar"
```
haus: it's -> its

shrug:
Ready to merge if @haus is happy?

haus:
@shrug i'm still reviewing. please hold 😄

src/puppetlabs/ezbake/core.clj
```diff
@@ -501,23 +541,112 @@ Dependency tree:
...
+                                resolved-dependencies))]
+    jar-file))
+
+(schema/defn extract-project-from-jar! :- schema/Str
+  "Extracts the project.clj file out of the given jar into the destination directory"
+  [jar :- File
+   destination-dir :- schema/Str]
+  (fs/mkdirs destination-dir)
+  (let [jar-file (JarFile. jar)
+        jar-entry (.getJarEntry jar-file "project.clj")
```
camlow325: The current work for the `additional-uberjars` is designed around retrieving repo artifacts and building an uberjar from them.  The jars deployed to repos for an artifact, including how we're handling the jruby-deps ones for Puppet Server, are typically not uberjars themselves.  They are just jars with compiled elements from their immediate project but with additional metadata indicating their upstream dependencies.  What we want to package into ezbake, though, are uberjars built from the immediate project jars which have been fetched from the repo.

For Clojure-based projects, building the uberjar from the project.clj is helpful because it can include any special packaging requirements, e.g., the uberjar-exclusions that we want for JRuby uberjars for Puppet Server.  It probably would be possible to extend this to build uberjars for pure-Java projects, e.g., using the pom file from the jar.  Not sure if this could be done with leiningen, though, or if we'd have to use Maven instead?  I definitely didn't want to try to get into that at this point until we had a specific use case for it.

Another reasonable extension to all of this at some point may be a way to, instead specifying coordinates for repo artifacts to build the uberjar from, provide a fully-qualified URL to some location where pre-built uberjars may have been distributed.  That could be pretty general-purpose (non-Clojure/Java specific).  That might also be nice for the JRuby uberjars at some point as well but I didn't want to force us to work through all of the delivery logistics for that at this point.

resources/puppetlabs/lein-ezbake/template/global/install.sh.erb
```diff
@@ -138,6 +138,12 @@ function task_install {
     install -m 0644 ext/ezbake.manifest "${DESTDIR}${app_prefix}"
     install -d -m 0755 "${DESTDIR}${projconfdir}/conf.d"
 
+<% if EZBake::Config[:additional_uberjars]  -%>
+  <% EZBake::Config[:additional_uberjars].each do |jar| -%>
+    install -m 0644 <%= jar %> "${DESTDIR}${app_prefix}"
```
jpinsonault: Seems reasonable

src/puppetlabs/ezbake/core.clj
```diff
@@ -501,23 +541,112 @@ Dependency tree:
...
+       :uberjar-name (:uberjar-name lein-project)
+       :additional-uberjars (mapv (fn [filename] {:uberjar filename}) additional-uberjars)
+       :is-pe-build (format "%s" (= (get-local-ezbake-var lein-project :build-type "foss") "pe"))
+       :repo-name (format "%s" (get-local-ezbake-var lein-project :repo-target ""))})))
+
+(schema/defn get-additional-uberjars
+  "Returns the list of additional uberjar dependencies from the given lein project"
+  [lein-project]
+  (when-let [dependencies-vector (get-in lein-project [:lein-ezbake :additional-uberjars])]
+    dependencies-vector))
```
jpinsonault: No I think you're right

src/puppetlabs/ezbake/core.clj
rlinehan: Cool, this makes sense. I'm alright with how it is now, and I think so long as the docs mention that it's using the `project.clj` (which they are), then this is all good.

src/puppetlabs/ezbake/core.clj
```diff
@@ -1,19 +1,21 @@
...
             [leiningen.deploy :as deploy]
+            [leiningen.core.classpath :as lein-classpath]
             [leiningen.uberjar :as uberjar]
             [schema.core :as schema]
             [schema.utils :as schema-utils]
             [puppetlabs.ezbake.dependency-utils :as deputils]
             [puppetlabs.ezbake.exec :as exec]
-            [puppetlabs.config.typesafe :as ts]))
+            [puppetlabs.config.typesafe :as ts]
+            [leiningen.core.project :as project]))
```
haus: looks like the list is pretty sorted, so this should go with the other lein requires

src/puppetlabs/ezbake/core.clj
```diff
@@ -173,31 +184,57 @@
...
+                        :dependencies (deputils/generate-dependency-tree-string (:lein-project uberjar-info))})
+        template-map
+        {:ezbake-version (get-ezbake-version lein-project)
+         :package-group (:group lein-project)
+         :package-name (:name lein-project)
+         :package-version (:version lein-project)
+         :manifest-string (deputils/generate-manifest-string lein-project)
+         :dependency-tree (deputils/generate-dependency-tree-string lein-project)
+         :additional-uberjar-info uberjar-info
+         :has-additional-uberjars (not (empty? additional-uberjars-info))}]
```
haus: `not-empty` instead? (https://clojuredocs.org/clojure.core/not-empty)

src/puppetlabs/ezbake/core.clj
```diff
@@ -501,23 +541,112 @@ Dependency tree:
...
+       :repo-name (format "%s" (get-local-ezbake-var lein-project :repo-target ""))})))
+
+(schema/defn get-additional-uberjars
+  "Returns the list of additional uberjar dependencies from the given lein project"
+  [lein-project]
+  (when-let [dependencies-vector (get-in lein-project [:lein-ezbake :additional-uberjars])]
+    dependencies-vector))
+
+(schema/defn resolve-dependency! :- File
+  "Resolves a single dependency and returns a File pointing to it's jar.
```
haus: it's => its

README.md
```diff
@@ -140,6 +140,44 @@ above; although it is conceivable that such a composite project may define its
...
+`cli-defaults.sh` is intended to be a file from which default shell script
+variables can be defined by a project, and be loaded by ezbake's `cli-app`
+script. That is, the script which is run when commands like
+`service my-project start` are called.
+
+This differs from the `default` files, which also get loaded by `cli-app`, in
+that a project has no control over what goes into `default`. `cli-defaults.sh`
+is under the control of your project.
+
+For `cli-defaults.sh` to be used, it simply needs to exist at
```
haus: What happens if more than one project in a composite has one of these files?

camlow325: I haven't tried that out but my guess is that the behavior would be similar to what happens with other resources with the same location in an dependency jar as in the downstream composite jar.  In those cases, I've observed that the downstream, composite one "wins" (is the one that gets bundled in the package).  If that is the behavior, does that seem reasonable to you at least for now?  I'd like to avoid having to complicate the logic by aggregating multiple possible files from each of the component jars and sourcing them all in some order from the startup scripts, etc.  Would be good to at least know what would happen in this situation, though - e.g., that the presence of multiple files in the different jars wouldn't cause an ezbake build to fail.

haus: That seems fine, it might be good to note that in the doc then.

camlow325: Hmmm, I tried this out with pe-puppetserver as my composite project and it didn't work the way I had been thinking it would.  I tried putting different versions of `./resources/ext/cli_defaults/cli-defaults.sh.erb` file into a pe-puppet-server-extensions and a pe-puppetserver jar and the ezbake stage from pe-puppetserver picked up the pe-puppet-server-extensions one, not the pe-puppetserver one.  I saw the same thing for other resources, e.g., for a cli app I stored under ext/cli.

Maybe ezbake doesn't automatically look at the resources in the immediate project's jar - at least unless you self reference the immediate project as a dependency [like we do in OSS puppetserver](https://github.com/puppetlabs/puppetserver/blob/2.7.2/project.clj#L143)?

jpinsonault:
pushed some commits to address feedback

README.md
camlow325: I think the behavior I'm describing where an ezbake build doesn't include a project's own resources unless the project is self-referenced in its own ezbake config is covered in [EZ-35](https://tickets.puppetlabs.com/browse/EZ-35).  I'm fine continuing to live with this limitation until/unless we have a pressing need to reconcile conflicting cli-defaults files in the parent composite project at some point.

camlow325:
I think @jpinsonault's latest changes have addressed all of the current feedback.  @haus, @rlinehan - any last objections before we merge this?

haus:
no, i'm good

rlinehan:
:shipit:

camlow325:
@jpinsonault Sorry, missed this earlier, but it looks like with the changes we merged that the "Additional Uberjars" and "Additional uberjar dependencies:" labels are duplicated in the ezbake.manifest file when ezbake is building more than 1 additional uberjar.  Maybe that happened after a later commit was added to the PR since I don't remember seeing that initially.

For my Puppet Server test builds, I see:

```
This package was built by the Puppet Labs packaging system.

EZBake version: 1.3.0
Release package: puppetlabs/puppetserver (5.0.0-master-SNAPSHOT)
Bundled packages: puppetlabs/puppetserver 5.0.0-master-SNAPSHOT,org.clojure/clojure 1.8.0,puppetlabs/puppetserver 5.0.0-master-SNAPSHOT,puppetlabs/trapperkeeper-webserver-jetty9 2.0.0
Additional Uberjars:
Additional Uberjars:
jruby-9k.jar
jruby-1_7.jar

...

Additional uberjar dependencies:

Additional uberjar dependencies:

jruby-9k.jar:

[org.jruby/jruby-core "9.1.8.0" :exclusions [[joda-time]]]
...
```

Probably would be good to fix this up later on, not a big deal.  Could file a separate ticket for that.  I am curious if you see the same thing.

rlinehan:
I'm also seeing this when I built a package off of https://github.com/puppetlabs/puppetserver/pull/1378. The ezbake.manfiest for that is at http://builds.puppetlabs.lan/puppetserver/5.0.0.master.SNAPSHOT.2017.05.04T1407/artifacts/5.0.0.master.SNAPSHOT.2017.05.04T1407.ezbake.manifest

jpinsonault:
ticketed, https://tickets.puppetlabs.com/browse/EZ-109

PR #417, opened 2017-05-11
(EZ-110) Fix ezbake manifest generation for dependency overrides

camlow325:
Previously, the additional uberjar functionality appended the immediate
project jar as a dependency of itself in order for its content to be
generated into the uberjar.  This logic, however, did not work when the
project had any dependency overrides in the uberjar profile.  In that
case, none of the content from the immediate jar ended up in the uberjar
and the ezbake.manifest dependency info errantly listed info from the
main project's dependencies which did not end up in the uberjar.

In this PR, the immediate project's jar is appended directly into
the process of creating the uberjar - rather than as a self-referential
dependency.  This allows for the uberjar to properly retain content from
the main project jar even in the case of uberjar profile overrides.  The
main project's own artifact info is also added to the dependency listing
info directly, avoiding the appearance of errant transitive dependency
info in the output in cases where dependency exclusions are used.

Also, work was previously done to expand SNAPSHOT versions to the version of
the actual artifact previously deployed to a repository.  In the process
of expansion, any qualifying info on the coordinate, e.g., any
exclusions, was lost.  The resulting uberjar which was built would still
honor the qualifying info, however, the ezbake.manifest would errantly
list info that should have been excluded per the qualifying coordinate
info.

In this PR, the SNAPSHOT expansion logic only updates the dependency
version field, leaving all other trailing coordinate info in place.
This enables the ezbake.manifest to accurately reflect the content
generated into the uberjar.

### DIFF
--- a/src/puppetlabs/ezbake/dependency_utils.clj
+++ b/src/puppetlabs/ezbake/dependency_utils.clj
@@ -64,6 +64,18 @@
 (defn expand-snapshot-version
   ([lein-project coords] (expand-snapshot-version lein-project coords {}))
   ([lein-project coords options]
+   ;; For cases where the -SNAPSHOT is expanded to a full version, the version
+   ;; portion of the coordinate (second element in the vector) is replaced but
+   ;; all other qualifying metadata in the coordinate, e.g., any :exclusions
+   ;; which might be present, is preserved.
+   ;;
+   ;; For example, if the original coordinate has:
+   ;;
+   ;;   ['foo "0.0.1-SNAPSHOT" :exclusions ['bar]]
+   ;;
+   ;; .. then the result after expansion might have:
+   ;;
+   ;;   ['foo "0.0.1-20170501.100101-123" :exclusions ['bar]]
    (assoc coords 1 (get-full-snapshot-version lein-project coords options))))
 
 (defn expand-snapshot-versions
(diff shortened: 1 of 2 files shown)


camlow325:
@jpinsonault, @haus - would appreciate it if you could look at this one.

src/puppetlabs/ezbake/dependency_utils.clj
```diff
@@ -63,8 +63,8 @@
 
 (defn expand-snapshot-version
   ([lein-project coords] (expand-snapshot-version lein-project coords {}))
-  ([lein-project [dependency version :as coords] options]
-   [dependency (get-full-snapshot-version lein-project coords options)]))
+  ([lein-project coords options]
+   (assoc coords 1 (get-full-snapshot-version lein-project coords options))))
```
jpinsonault: This function could probably use a docstring, since it's a little non-intuitive that it's trying to preserve :exclusions

camlow325:
Thanks for the feedback, @jpinsonault.  I think I addressed it in f4c1bf8.

haus:
This seems gross and effective. A winning combination.

camlow325:
@haus Since I've addressed @jpinsonault's feedback and sounds like you're ok with this, would you mind merging?

PR #420, opened 2017-06-20
(RE-8457) Update ezbake to ship projects to puppet5

melissa:
This commit switches ezbake to use the 1.0.x branch of
puppetlabs/packaging, which contains all the updates we need to ship
packages to the new puppet5 and puppet5-nightly repos. This commit also
adds in the new variable `gpg_nonfinal_key`, which allows us to sign
nonfinal/development/beta/nightly builds with a separate key.

### DIFF
--- a/resources/puppetlabs/lein-ezbake/staging-templates/project_data.yaml.mustache
+++ b/resources/puppetlabs/lein-ezbake/staging-templates/project_data.yaml.mustache
@@ -65,9 +65,11 @@ tar_excludes:
   - .gitignore
 build_pe: {{{is-pe-build}}}
 apt_repo_name: {{{repo-name}}}
-apt_nonfinal_repo_name: {{{nonfinal-repo-name}}}
 yum_repo_name: {{{repo-name}}}
-yum_nonfinal_repo_name: {{{nonfinal-repo-name}}}
+repo_name: {{{new-repo-name}}}
+repo_link_target: {{{repo-link-name}}}
+nonfinal_repo_name: {{{nonfinal-repo-name}}}
+nonfinal_repo_link_target: {{{nonfinal-repo-link-name}}}
 gem_files:
 gem_require_path:
 gem_test_files:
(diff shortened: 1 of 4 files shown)


melissa:
Please don't merge this yet. We need to make sure this is in fact targeted at the correct branch. This change is specifically for projects targeting the new puppet5 repos. It will break if we attempt to use it with any branch of the packaging repo that doesn't have the updates currently in the 1.0.x branch

resources/puppetlabs/lein-ezbake/staging-templates/project_data.yaml.mustache
```diff
@@ -59,10 +59,10 @@ templates:
...
   - .gitignore
 build_pe: {{{is-pe-build}}}
-apt_repo_name: {{{repo-name}}}
-apt_nonfinal_repo_name: {{{nonfinal-repo-name}}}
-yum_repo_name: {{{repo-name}}}
-yum_nonfinal_repo_name: {{{nonfinal-repo-name}}}
+repo_name: {{{repo-name}}}
+repo_link_target: 'puppet'
+nonfinal_repo_name: {{{nonfinal-repo-name}}}
+nonfinal_repo_link_target: 'puppet-nightly'
```
melissa: I originally wasn't going to do this, but I'm thinking it's going to be a lot easier if we hard code these link targets in. Thoughts, @underscorgan . I know this is a public repo, and theoretically the public could be using it... but... also, this would be awesome to have

underscorgan: @melissa I was wondering if we could have these be overwritable with a parameter but maybe default to the hardcoded values?

melissa: That sounds like the best path forward. I'm  not positive how to do that with mustache, but I'll look into it

shrug:
Waiting on BKR-1118

PR #422, opened 2017-06-30
(MAINT) Prepare for 1.5.0 release

nwolfe:


project.clj
```diff
@@ -1,4 +1,4 @@
-(defproject puppetlabs/lein-ezbake "1.4.1-SNAPSHOT"
+(defproject puppetlabs/lein-ezbake "1.5.0-SNAPSHOT"
```
shrug: I think the jenkins release jobs will actually update this

nwolfe: I looked at the previous commits in the repo that prepared for releases and they do this.

The jenkins job will just strip off the `-SNAPSHOT`, do the release (so we get `1.5.0`), then update the string to `1.5.1-SNAPSHOT`.

nwolfe: See: https://github.com/puppetlabs/ezbake/commit/358c59e7945e38cabe6511360cb48d9c8d96b67c

nwolfe: And the job we ran: https://jenkins-enterprise.delivery.puppetlabs.net/view/All/job/enterprise_ezbake_release-clj_master/4/

nwolfe: And another example: https://github.com/puppetlabs/ezbake/commit/72ece8d1ab8481b9f4925714f33a031534e57dc4

shrug: Yep, thanks!

PR #426, opened 2017-07-11
(RE-8861) Add managed dependencies to dependency map

demophoon:
Before this commit we were only listing the dependencies which were
listed in the dependencies field which may not include everything. This
commit adds the managed dependencies to the list of dependencies to
consider when building out the dependency map.


demophoon:
@nwolfe Does this seem reasonable? This seems to solve at least getting dependencies from the primary uberjar. Example output from pe-puppetserver: https://gist.github.com/demophoon/38a760437cc6cae04efbce65e8db4dab

Mind this doesn't get the dependencies from any of the extra uberjars which are specified to be included with ezbake.

PR #429, opened 2017-07-27
(maint) Move doc installation to install.sh

underscorgan:
As we start exploring alternate packaging with ezbake, it's cleaner if
we let the install script put the docs in place as we do with all the
other files rather than installing those separately.


haus:
Isn't docdir defined slightly differently between debuntu and el?

On debuntu, the package docdir is `/usr/share/doc/${package_name}` while on el is it `/usr/share/doc/${package_name}-${package_version}`.

This PR puts it all under the el standard, which might be confusing for deb users?

underscorgan:
@haus Ahh, good point. I have an idea to work around that I think

underscorgan:
@haus this look better?

PR #432, opened 2017-08-03
(RE-8990) Start supporting FPM builds for ezbake

underscorgan:


underscorgan:
Still need to add docs.

resources/puppetlabs/lein-ezbake/staging-templates/project_data.yaml.mustache
```diff
@@ -36,12 +36,16 @@ templates:
     target: ext/{{{project}}}.logrotate-legacy.conf
   - source: ext/ezbake.tmpfiles.conf.erb
     target: ext/{{{project}}}.tmpfiles.conf
+  - ext/fpm.rb.erb
+  - ext/redhat/preinst.erb
+  - ext/redhat/postinst.erb
+  - ext/redhat/prerm.erb
+  - ext/redhat/postrm.erb
```
melissa: Does rpm support external pre/post scripts like this? From my experience, these have just lived in the spec file

i.e. `https://github.com/puppetlabs/ezbake/blob/master/resources/puppetlabs/lein-ezbake/template/global/ext/redhat/ezbake.spec.erb#L206-L223`
```
%pre
# Note: changes to this section of the spec may require synchronisation with the
# install.sh source based installation methodology.
#
# Add <%= EZBake::Config[:group] %> group
getent group <%= EZBake::Config[:group] %> > /dev/null || \
  groupadd -r <%= EZBake::Config[:group] %> || :
# Add <%= EZBake::Config[:user] %> user
if getent passwd <%= EZBake::Config[:user] %> > /dev/null; then
  usermod --gid <%= EZBake::Config[:group] %> --home %{_app_data} \
  --comment "<%= EZBake::Config[:project] %> daemon" <%= EZBake::Config[:user] %> || :
else
  useradd -r --gid <%= EZBake::Config[:group] %> --home %{_app_data} --shell $(which nologin) \
    --comment "<%= EZBake::Config[:project] %> daemon"  <%= EZBake::Config[:user] %> || :
fi
<% EZBake::Config[:redhat][:additional_preinst].each do |cmd| -%>
<%= cmd %>
<% end -%>
```

underscorgan: For RPM they live in the spec file, but FPM takes them as external files then uses those to generate the specs

melissa: Interesting. Does it use the existing spec file at all, or create an entirely new one with these external files?

underscorgan: It generates an entirely new one with these files and all the parameters that are passed to the fpm command. I'll grab a spec file and what the fpm command ends up looking like for deb/rpm to share here.

resources/puppetlabs/lein-ezbake/template/global/controller.sh.erb
```diff
@@ -0,0 +1,146 @@
...
+        DESTDIR=old_el bash install.sh sysv_init_redhat
+fi
+if [ ! -d old_sles ]; then
+        cp -r base old_sles
+        DESTDIR=old_sles bash install.sh sysv_init_suse
+fi
+
+# things are only different if we have docs, deb docs get
+# installed in an unversioned folder but rpm docs get installed
+# in a versioned folder.
```
melissa: Did you check on this with different versions of rpm/deb platforms we support? Is there any variation there, like Mike indicated there might be?

resources/puppetlabs/lein-ezbake/template/global/controller.sh.erb
```diff
@@ -0,0 +1,146 @@
...
+                echo "I have no idea what I'm doing with $build_os, teach me?" >&2
+                exit 1
+                ;;
+esac
+
+# bash will eat your spaces, so let's array. see http://mywiki.wooledge.org/BashFAQ/050 for more fun.
+params=("--user" "<%= EZBake::Config[:user] -%>" "--group" "<%= EZBake::Config[:group] -%>" "--chdir" "${dir}" "--realname" "<%= EZBake::Config[:real_name] -%>" "--debug" "--operating-system" "${os}" "--name" "<%= EZBake::Config[:project] -%>" "--package-version" "<%= EZBake::Config[:version] -%>" "--release" "<%= EZBake::Config[:release] -%>")
+if [ -n "${os_version}" ]; then params+=("--os-version" "${os_version}"); fi
+if [ -n "${os_dist}" ]; then params+=("--dist" "${os_dist}"); fi
+
```
e-gris: _Minor_ style comment: ${x} is rarely needed over $x. When I see braces used for variables it triggers an alert that something *special* is happening here. I has a dissapoint when there isn't.

resources/puppetlabs/lein-ezbake/template/global/controller.sh.erb
underscorgan: Yeah, debs are in unversioned dirs and rpms at least as far back as el6 use versioned dirs which is handled in https://github.com/puppetlabs/ezbake/pull/429

resources/puppetlabs/lein-ezbake/template/global/controller.sh.erb
underscorgan: ah, yeah, I can get rid of those. Those are relics of trying to figure out why bash was eating all of the spaces in my array

underscorgan:
fpm command for building debs:
```
fpm  --deb-build-depends cdbs --deb-build-depends bc --deb-build-depends mawk --deb-build-depends lsb-release --deb-build-depends "ruby | ruby-interpreter" --deb-priority optional --category utils --name puppetserver --depends openjdk-8-jre-headless --depends bash --depends net-tools --depends procps --depends 'puppet-agent (>= 4.99.0)' --before-install ext/debian/preinst --after-install ext/debian/postinst --before-remove ext/debian/prerm --after-remove ext/debian/postrm --force --chdir systemd_deb  --version 5.0.1.master --iteration 0.1SNAPSHOT.2017.08.04T1812stretch --vendor 'Puppet Labs <info@puppetlabs.com>' --maintainer 'Puppet Labs <info@puppetlabs.com>' --license 'ASL 2.0' --url http://puppet.com --architecture all --output-type deb --input-type dir etc lib opt usr var
```

underscorgan:
fpm command for building rpms
```
FPM_EDITOR="sed -i 's/%dir %attr(-/%attr(-/'" fpm  --rpm-rpmbuild-define '_app_logdir /var/log/puppetlabs/puppetserver' --rpm-rpmbuild-define '_app_rundir /var/log/puppetlabs/puppetserver' --rpm-rpmbuild-define '_app_prefix /opt/puppetlabs/server/apps/puppetserver' --rpm-rpmbuild-define '_app_data /opt/puppetlabs/server/data/puppetserver' --rpm-rpmbuild-define '_with_sysvinit 0' --rpm-rpmbuild-define '_with_systemd 1' --rpm-rpmbuild-define '_old_sles 0' --rpm-rpmbuild-define '_systemd_el 1' --rpm-rpmbuild
-define '_systemd_sles 0' --rpm-rpmbuild-define '_old_el 0' --rpm-rpmbuild-define '_sysconfdir /etc' --rpm-rpmbuild-define '_prefix /opt/puppetlabs/server/apps/puppetserver' --rpm-rpmbuild-define '_rundir /var/run' --rpm-rpmbuild-define '__jar_repack 0' -
-depends systemd --config-files /etc/puppetlabs/puppetserver --config-files /etc/sysconfig/puppetserver --directories /var/log/puppetlabs/puppetserver --directories /etc/puppetlabs/puppetserver --directories /var/run/puppetlabs/puppetserver --rpm-auto-add
-exclude-directories /etc/puppetlabs --rpm-auto-add-exclude-directories /opt/puppetlabs/bin --rpm-auto-add-exclude-directories /opt/puppetlabs/server --rpm-auto-add-exclude-directories /opt/puppetlabs/server/apps --rpm-auto-add-exclude-directories /opt/pu
ppetlabs/server/bin --rpm-auto-add-exclude-directories /opt/puppetlabs/server/data --rpm-auto-add-exclude-directories /usr/lib/systemd --rpm-auto-add-exclude-directories /usr/lib/systemd/system --rpm-auto-add-exclude-directories /etc/init.d --rpm-auto-add
-exclude-directories /etc/rc.d --rpm-auto-add-exclude-directories /etc/logrotate.d --rpm-auto-add-exclude-directories /etc/rc.d/init.d --rpm-auto-add-exclude-directories /usr/lib/tmpfiles.d --rpm-auto-add-exclude-directories /var/log/puppetlabs --rpm-auto-add-exclude-directories /var/run/puppetlabs --rpm-attr 750,puppet,puppet:/etc/puppetlabs/puppetserver --rpm-attr 700,puppet,puppet:/var/log/puppetlabs/puppetserver --rpm-attr -,puppet,puppet:/opt/puppetlabs/server/data/puppetserver --rpm-attr 755,puppet,puppet:/var/run/puppetlabs/puppetserver --edit --category 'System Environment/Daemons' --name puppetserver --depends java-1.8.0-openjdk-headless --depends ca-certificates --depends bash --depends net-tools --depends /usr/bin/which --depends procps --depen
ds 'puppet-agent >= 4.99.0' --before-install ext/redhat/preinst --after-install ext/redhat/postinst --before-remove ext/redhat/prerm --after-remove ext/redhat/postrm --force --chdir systemd_el  --rpm-rpmbuild-define 'rpmversion 5.0.1.master' --rpm-dist el
7 --rpm-auto-add-directories --rpm-auto-add-exclude-directories /opt/puppetlabs --version 5.0.1.master --iteration 0.1SNAPSHOT.2017.08.04T1812 --vendor 'Puppet Labs <info@puppetlabs.com>' --maintainer 'Puppet Labs <info@puppetlabs.com>' --license 'ASL 2.0
' --url http://puppet.com --architecture all --output-type rpm --input-type dir etc opt usr var
```

underscorgan:
the auto-file additions make this a little bonkers, but here's a spec for puppetserver https://gist.github.com/underscorgan/dd5bf87bca6df7fe0eccfc395d9c2b28

resources/puppetlabs/lein-ezbake/template/global/ext/fpm.rb.erb
```diff
@@ -0,0 +1,376 @@
+#!/bin/ruby
```
underscorgan: I think I created this as an erb when I was starting with the scripting, thinking I might end up using some of the ezbake variable interpolation, but those all ended up in the controller, so I can make this not a template.

Ah, yeah. I should've used `#!/usr/bin/env ruby` for more portability.

resources/puppetlabs/lein-ezbake/template/global/ext/fpm.rb.erb
```diff
@@ -0,0 +1,376 @@
...
+
+require 'open3'
+require 'optparse'
```
e-gris: Have you ever used docopt? It's ReallyCool.

underscorgan: whoa, no. That is very cool

melissa:
I'm curious what the fpm generated spec files looks like in comparison to ours. Is there a way to see that?

underscorgan:
ezbake spec: https://gist.github.com/underscorgan/e7909e67709e094b8e04a41e93e57e49

src/puppetlabs/ezbake/core.clj
```diff
@@ -442,6 +442,25 @@ Additional uberjar dependencies:
...
+            (str/replace lein-version #"-SNAPSHOT|-" { "-SNAPSHOT" "" "-" "."}))
+    lein-version))
+
+(defn generate-package-release-from-version
+  [lein-version]
+  {:pre [(string? lein-version)]
+   :post [(string? %)]}
+  (if (.endsWith lein-version "-SNAPSHOT")
+    (format "%s.%s"
+            (str/replace lein-version #".*-" "0.1")
```
nwolfe: I'm not exactly sure the context of this method, but this will always be `"0.1"`? How fortunate!

underscorgan: EZBake doesn't support different release numbers (you don't end up with like puppetserver 5.0.0-1 and puppetserver 5.0.0-2), and packaging prepends SNAPSHOT with 0.<release>SNAPSHOT to help with correct package versioning. https://github.com/puppetlabs/packaging/pull/393/files

melissa: for CPR-429, should we add the codename to the version or to the package name? I think in vanagon, it's appended on as a part of the release field (i.e., `puppet-agent_1.9.1-1xenial_i386.deb`), if that helps

melissa: I'd love to piggy back the package name change onto this PR, but if not, that's fine

resources/puppetlabs/lein-ezbake/template/global/ext/fpm.rb
```diff
@@ -0,0 +1,376 @@
...
+  fpm_opts << "--rpm-attr -,#{options.user},#{options.group}:#{options.app_data}"
+  fpm_opts << "--rpm-attr 755,#{options.user},#{options.group}:#{options.app_rundir}"
+
+  fpm_opts << "--edit"
+  fpm_opts << "--category 'System Environment/Daemons'"
+  termini_opts << "--category 'Development/Libraries'"
+#deb specific options
+elsif options.output_type == 'deb'
+  if options.dist != "#{options.operating_system}#{options.os_version}"
+    options.release = "#{options.release}#{options.dist}"
```
underscorgan: @melissa we append the dist to the release here!

resources/puppetlabs/lein-ezbake/template/global/ext/fpm.rb
```diff
@@ -0,0 +1,376 @@
...
+  fpm_opts << "--rpm-rpmbuild-define '_old_sles #{options.old_sles}'"
+  fpm_opts << "--rpm-rpmbuild-define '_systemd_el #{options.systemd_el}'"
+  fpm_opts << "--rpm-rpmbuild-define '_systemd_sles #{options.systemd_sles}'"
+  fpm_opts << "--rpm-rpmbuild-define '_old_el #{options.old_el}'"
+  fpm_opts << "--rpm-rpmbuild-define '_sysconfdir /etc'"
+  fpm_opts << "--rpm-rpmbuild-define '_prefix #{options.app_prefix}'"
+  fpm_opts << "--rpm-rpmbuild-define '_rundir /var/run'"
+  fpm_opts << "--rpm-rpmbuild-define '__jar_repack 0'"
+
+  shared_opts << "--rpm-dist #{options.dist}"
```
underscorgan: and for rpm the appending is done via the `--rpm-dist` option

resources/puppetlabs/lein-ezbake/template/global/ext/fpm.rb
melissa: OH YAY! Thank you <3

README.md
```diff
@@ -234,6 +234,25 @@ version of PE you're building a package for.  e.g.:
...
+on your local machine using [FPM](https://github.com/jordansissel/fpm). To
+build successfully you'll need the FPM gem installed. You'll also need java
+and leiningen. To build RPMs you'll want to be on some sort of RPM-based system
+as you need RPM build tools. To build RPMs for SLES you'll need to have 
+the systemd-rpm-macros rpm installed. Building .debs doesn't require anything
+special.
+
+Packages will end up in the output directory. RPM targets can be overwritten
+by setting the `MOCK` environment variable and deb targets can be overwritten
+by setting the `COW` environment variable.
```
melissa: Is there a certain format expected here? A space or coma separated list, for example?

resources/puppetlabs/lein-ezbake/template/global/controller.sh.erb
```diff
@@ -0,0 +1,146 @@
...
+                cp -r base systemd_deb
+                DESTDIR=systemd_deb bash install.sh systemd_deb
+        fi
+        if [ ! -d sysvinit_deb ]; then
+                cp -r base sysvinit_deb
+                DESTDIR=sysvinit_deb bash install sysv_init_deb
+        fi
+fi
+
+case $build_os in
```
melissa: Hmm, so I can see why you did this, but is there a way to programmatically do this? I'm just hesitating because I'd rather not have to update this every time we add a new platform

resources/puppetlabs/lein-ezbake/template/global/controller.sh.erb
```diff
@@ -0,0 +1,146 @@
...
+                dir=systemd_deb
+                os_dist=yakkety
+                ;;
+        *)
+                echo "I have no idea what I'm doing with $build_os, teach me?" >&2
+                exit 1
+                ;;
+esac
+
+# bash will eat your spaces, so let's array. see http://mywiki.wooledge.org/BashFAQ/050 for more fun.
```
melissa: oh bash

resources/puppetlabs/lein-ezbake/template/global/controller.sh.erb
```diff
@@ -0,0 +1,146 @@
...
+DESTDIR=termini bash install.sh termini
+params+=('--build-termini')
+<% end -%>
+
+<% if EZBake::Config[:logrotate_enabled]-%>
+DESTDIR=systemd_el bash install.sh logrotate
+DESTDIR=systemd_deb bash install.sh logrotate
+DESTDIR=sysvinit_deb bash install.sh logrotate
+DESTDIR=old_el bash install.sh logrotate_legacy
+DESTDIR=old_sles bash install.sh logrotate_legacy
```
melissa: Hmm, maybe I'm missing something, but it just looks like you're overwriting `DESTDIR` with these different calls to `install.sh`... I'm not sure what `install.sh` will return, but the last set will win. Is that intentional?

resources/puppetlabs/lein-ezbake/template/global/controller.sh.erb
```diff
@@ -0,0 +1,146 @@
...
+params+=('--logrotate')
+<% end -%>
+
+<% if EZBake::Config[:is_pe_build] -%>
+params+=('--enterprise-build')
+<% end -%>
+
+# pull in rpm dependencies
+if [[ "$build_os" =~ el-?7 || "$build_os" =~ el-?6 || "$build_os" =~ sles || "$build_os" =~ fedora ]]; then
```
melissa: Any reason we specify either el6 or el7, and not just all el in general? We'll have to update this when we add in el8. I suppose because you have to match against `el`, and that's a relatively flimsy match criteria...

resources/puppetlabs/lein-ezbake/template/global/controller.sh.erb
```diff
@@ -0,0 +1,146 @@
...
+        <% end -%>
+# if we aren't an rpm, pull in deb dependencies
+else
+        <%EZBake::Config[:debian][:additional_dependencies].each do |dep| -%>
+                params+=("--additional-dependency")
+                params+=("<%= dep -%>")
+        <% end -%>
+fi
+
+ruby $PWD/ext/fpm.rb "${params[@]}"
```
melissa: Should we fully quality the path to `ruby`?

resources/puppetlabs/lein-ezbake/template/global/ext/fpm.rb
```diff
@@ -0,0 +1,376 @@
+#!/usr/bin/env ruby
```
melissa: Is this a script you put together, or is it a generic fpm script that you've made lots of modifications to? I'm wondering about updating it for newer versions of fpm, if that's something we'll need to do in the future. If it's totally a custom one, and we don't need to update it based on an upstream script, just ignore me

underscorgan: This is a totally custom script, it's basically converting the existing spec and control/rules/etc files into fpm commands.

resources/puppetlabs/lein-ezbake/template/global/ext/redhat/postrm.erb
```diff
@@ -0,0 +1,18 @@
+%if %{_systemd_el}
+%systemd_postun_with_restart <%= EZBake::Config[:project] %>.service
+%endif
+%if %{_systemd_sles}
+%service_del_postun <%= EZBake::Config[:project] %>.service
+%endif
+%if %{_with_sysvinit}
+# Remove the rundir if this is an uninstall (as opposed to an upgrade)...
+if [ "$1" = "0" ]; then
+    rm -rf %{_app_rundir} || :
```
demophoon: I might be missing something, but what does the `|| :` do here?

resources/puppetlabs/lein-ezbake/template/global/controller.sh.erb
e-gris: shell commands optionally take a set of shell environments definitions before the commands:

FOO=bar shell-command <arguments>

These lines are calling bash with the DESTDIR environment variable set differently for each invocation.

melissa: Oooh, duh, thank you for the clarification, Eric!

resources/puppetlabs/lein-ezbake/template/global/controller.sh.erb
underscorgan: yeah, that was my thinking :\

resources/puppetlabs/lein-ezbake/template/global/ext/redhat/postrm.erb
underscorgan: `:` is aliased to `true` so it won't fail if something goes wrong with the `rm -rf`

resources/puppetlabs/lein-ezbake/template/global/controller.sh.erb
underscorgan: Probably not. Since this will be able to run in a bunch of places which could have different ruby setups we should probably trust the PATH

resources/puppetlabs/lein-ezbake/template/global/controller.sh.erb
demophoon: It might be good to pull these out into helper functions like `is_supported_el?` and `is_supported_fedora?` so that adding new platforms isn't a big chore for us down the road. Maybe potentially in a helper bash script that we can render in this erb template that contains all of these functions.

resources/puppetlabs/lein-ezbake/template/global/controller.sh.erb
underscorgan: It would be great if it were possible to do this programmatically, but I'm not sure how we could make that work. Hmm, we might be able to encode this information in packaging and pass that as environment variables, then at least we'd only need to update it in one place. The directories will still need to be set in here since packaging shouldn't know about those details, and if we update the FPM script to use 2-digit sles versions that would probably help (those I found in a chart online that I'm failing to find again right now), though built-in macros might do something strange if the version isn't how it's expecting it to be

nwolfe: What calls this script?

Maybe these could all be inputs, rather than just the OS name and then we derive these values from that?

```
./controller.sh <os> <dir> <os_version optional> <os_dist optional>

File: resources/puppetlabs/lein-ezbake/template/global/tasks/build.rake [hunk 17]
```diff
@@ -0,0 +1,33 @@
...
+          FileUtils.mv(Dir.glob("*.rpm"), "#{nested_output}/#{mock}")
+        end
+        Pkg::Config.cows.split(" ").each do |cow|
+          FileUtils.mkdir("#{nested_output}/#{cow}") unless File.directory?("#{nested_output}/#{cow}")
+          `bash controller.sh #{cow}`
+          FileUtils.mv(Dir.glob("*.deb"), "#{nested_output}/#{cow}")
+        end
+      end
+    end
+	end
```
melissa: ARG! FORMATTING!

melissa:
This looks so good!

resources/puppetlabs/lein-ezbake/template/global/ext/redhat/postrm.erb
e-gris: rm -f never complains and always returns 0 so the '|| :' can be dropped without losing any function.

resources/puppetlabs/lein-ezbake/template/global/tasks/build.rake
underscorgan: oh no! i had a mix of spaces and a tab here, so it looked right in my terminal. Fixing.

resources/puppetlabs/lein-ezbake/template/global/controller.sh.erb
```diff
@@ -0,0 +1,135 @@
...
+                        dir="$BASEPATH/systemd_el"
+                else
+                        dir="$BASEPATH/old_sles"
+                fi
+                ;;
+        *)
+                echo "I have no idea what I'm doing with $os, teach me?" >&2
+                exit 1
+                ;;
+esac
```
underscorgan: @melissa is this case statement less awful? We can easily pass in the version and os from the rake file where we call this, but I sorta feel like since the directories are all created in this script, making something else know what they are feels messy.

resources/puppetlabs/lein-ezbake/template/global/controller.sh.erb
```diff
@@ -131,7 +117,7 @@ params+=('--enterprise-build')
 <% end -%>
 
 # pull in rpm dependencies
-if [[ "$build_os" =~ el-?7 || "$build_os" =~ el-?6 || "$build_os" =~ sles || "$build_os" =~ fedora ]]; then
+if [[ "$os" = 'el' || "$os" = 'sles' || "$os" = 'fedora' ]]; then
```
underscorgan: @demophoon does this work for you as simpler than hardcoding supported platform versions? This should only need updates if we add a new OS all together.

README.md
underscorgan: ah, yeah. These should be a space-separated list, same as the packaging environment variable overrides. I'll add a comment here

resources/puppetlabs/lein-ezbake/template/global/controller.sh.erb
demophoon: This looks much better! Less for us to have to remember to update in the future.

underscorgan:
also FYI this is what the output dir for the local builds looks like
```
[morgan@fpm puppetdb]$ find output/
output/
output/el
output/el/6
output/el/6/puppetdb-5.1.0-0.1SNAPSHOT.2017.08.18T1818.el6.noarch.rpm
output/el/6/puppetdb-termini-5.1.0-0.1SNAPSHOT.2017.08.18T1818.el6.noarch.rpm
output/el/7
output/el/7/puppetdb-5.1.0-0.1SNAPSHOT.2017.08.18T1818.el7.noarch.rpm
output/el/7/puppetdb-termini-5.1.0-0.1SNAPSHOT.2017.08.18T1818.el7.noarch.rpm
output/sles
output/sles/12
output/sles/12/puppetdb-5.1.0-0.1SNAPSHOT.2017.08.18T1818.sles12.noarch.rpm
output/sles/12/puppetdb-termini-5.1.0-0.1SNAPSHOT.2017.08.18T1818.sles12.noarch.rpm
output/deb
output/deb/jessie
output/deb/jessie/puppetdb_5.1.0-0.1SNAPSHOT.2017.08.18T1818jessie_all.deb
output/deb/jessie/puppetdb-termini_5.1.0-0.1SNAPSHOT.2017.08.18T1818jessie_all.deb
output/deb/xenial
output/deb/xenial/puppetdb_5.1.0-0.1SNAPSHOT.2017.08.18T1818xenial_all.deb
output/deb/xenial/puppetdb-termini_5.1.0-0.1SNAPSHOT.2017.08.18T1818xenial_all.deb
output/deb/stretch
output/deb/stretch/puppetdb_5.1.0-0.1SNAPSHOT.2017.08.18T1818stretch_all.deb
output/deb/stretch/puppetdb-termini_5.1.0-0.1SNAPSHOT.2017.08.18T1818stretch_all.deb
```

resources/puppetlabs/lein-ezbake/template/global/controller.sh.erb
melissa: I think this is the best we can do given what we're working with. I think it's definitely reasonable

resources/puppetlabs/lein-ezbake/template/global/controller.sh.erb
```diff
@@ -0,0 +1,135 @@
+#!/bin/bash
+
+################
+# Usage: controller.sh <os>
```
melissa: does this need to be updated to `controller.sh <os> <ver>`?

resources/puppetlabs/lein-ezbake/template/global/tasks/build.rake
```diff
@@ -0,0 +1,41 @@
...
+          `bash controller.sh #{os} #{ver}`
+          FileUtils.cp(Dir.glob("*#{os}#{ver}*.rpm"), "#{nested_output}/#{platform_path}")
+        end
+        Pkg::Config.cows.split(" ").each do |cow|
+          platform = cow.split('-')[1..-2].join('-')
+          platform_path = "deb/#{platform}"
+          FileUtils.mkdir_p("#{nested_output}/#{platform_path}") unless File.directory?("#{nested_output}/#{platform_path}")
+          # there's no differences in packaging for deb vs ubuntu so picking debian
+          # if that changes we'll need to fix that
+          `bash controller.sh debian #{platform}`
```
melissa: is this always going to be `debian`? Don't we have some `ubuntu` cows as well?

melissa: oops, I should have read the comment you have on line 33. IGNORE ME!

underscorgan: We have no packaging differences between debian and ubuntu so I just picked one and left a comment above

resources/puppetlabs/lein-ezbake/template/global/controller.sh.erb
underscorgan: good call!

resources/puppetlabs/lein-ezbake/template/global/controller.sh.erb
```diff
@@ -0,0 +1,135 @@
...
+# and currently supports el6, el7, sles11, sles12, debian jessie,
+# debian stretch, ubuntu trusty, ubuntu xenial, and ubuntu yakkety
+################
+build_os=$1
+build_ver=$2
+
+BASEPATH='pkg_artifacts'
+mkdir $BASEPATH
+if [ ! -d "$BASEPATH/base" ]; then
+        DESTDIR="$BASEPATH/base" bash install.sh install_redhat
```
melissa: is there an `install_debian` method here? Should we distinguish? Or is this fine for all cases?

melissa: it looks like `DESTDIR="$BASEPATH/base_deb" bash install.sh install_deb` is only run if there are docs?

underscorgan: The only difference between install_deb and install_redhat is if there are docs (whether or not to be in a versioned directory), so if there aren't any docs installed with the package there's no need to run it again

melissa: Oh that's awesome!

melissa:
I haven't reviewed `fpm.rb` very closely, but overall, I'm definitely 👍 on getting this in. I think I had a few minor nits that stuck out, but other than that it looks good.

I'm also wondering if now is an appripriate time to update the output dir to stage packages in the tag directories (i.e., `el-7-x86_64`).  I'm definitely okay with that getting in later, but if we make that decision, we have to open a new ticket for that work

resources/puppetlabs/lein-ezbake/template/global/tasks/build.rake
```diff
@@ -0,0 +1,40 @@
...
+      Pkg::Config.cows.split(" ").each do |cow|
+        platform = cow.split('-')[1..-2].join('-')
+        platform_path = "deb/#{platform}"
+        FileUtils.mkdir_p("#{pkg_path}/#{platform_path}") unless File.directory?("#{pkg_path}/#{platform_path}")
+        # there's no differences in packaging for deb vs ubuntu so picking debian
+        # if that changes we'll need to fix that
+        `bash controller.sh debian #{platform}`
+        FileUtils.cp(Dir.glob("*#{platform}*.deb"), "#{pkg_path}/#{platform_path}")
+      end
+      FileUtils.cp_r(pkg_path, nested_output)
```
melissa: We're probably going to need to include the repo name to maintain backward compatibility (i.e., `PC1`, `puppet5`)

underscorgan: yeah, I think I should be able to grab those from packaging

MosesMendoza:
Cool work. I'm curious what the broader workflow intention/direction is for this PR? I checked out RE-8990 to gather some background, but it wasn't entirely clear. I did have pause from:
> it's going to be less of a direct drop-in replacement for uber_build, more a replacement for the rpm spec files and deb control/rules/etc files that already live in EZBake

Just to make sure I understand - we're still going to support the existing workflow that allows me to  make a package from my workstation without needing build tools like rpmbuild or mock installed, right? The part of that I find important is it allows me to create packages of ie pe-puppetserver from my local checkout on my workstation that I can plug the packages into local beaker acceptance runs. For example, I built a custom package http://builds.delivery.puppetlabs.net/pe-puppetserver/2017.2.0.24.SNAPSHOT.2017.08.18T1604 for PR https://github.com/puppetlabs/pe-puppetserver/pull/215 in order to run the pe-puppet-server-extensions file sync acceptance tests on it locally without needing my PR merged.

underscorgan:
Hey @MosesMendoza -- as far as broader workflow and direction this work is intended to be useful in a number of ways. If you happen to be on a machine/vm/etc where rpm-build is available this will allow you to perform ezbake builds locally. They can also be performed locally using docker. Finally, I've been working on setting up jenkins jobs that can be triggered by other jenkins jobs or from the command line (using curl to trigger an existing job rather than creating a new dynamic job by passing a blob of xml). While the jenkins job would let you build an arbitrary fork/branch of an ezbake project I still need to work out something more like the existing `lein with-profile ezbake ezbake build` since I've found the jenkins job very tricky for building with a local branch for ezbake. The end-goal for this work is getting builds off of the rpm-builder and deb-builder machines since those are slow and overworked, instead moving builds into containers that will work with the current state of CI (the performance improvements are vast).

As for 
>it's going to be less of a direct drop-in replacement for uber_build, more a replacement for the rpm spec files and deb control/rules/etc files that already live in EZBake

when I was initially scoping this work I was wondering if the FPM build process was going to be generic enough where it could end up replacing the existing uber_build steps in packaging, but once I started getting down into the work I found FPM was more a tool that let me model the spec and control/rules/etc files that already existed in ezbake, so it made sense to be making these changes in ezbake rather than changing the more generic and widely used methods in packaging.

resources/puppetlabs/lein-ezbake/template/global/tasks/build.rake
underscorgan: This is added in 772a416b615ed3698e868ca893da29ae3225efb0

underscorgan:
New contents of output:
```
[morgan@fpm puppetserver]$ find output -type f
output/el/6/PC1/x86_64/puppetserver-5.0.1.master-0.1SNAPSHOT.2017.08.22T1658.el6.noarch.rpm
output/el/6/PC1/i386/puppetserver-5.0.1.master-0.1SNAPSHOT.2017.08.22T1658.el6.noarch.rpm
output/el/7/PC1/x86_64/puppetserver-5.0.1.master-0.1SNAPSHOT.2017.08.22T1658.el7.noarch.rpm
output/el/7/PC1/i386/puppetserver-5.0.1.master-0.1SNAPSHOT.2017.08.22T1658.el7.noarch.rpm
output/sles/12/PC1/x86_64/puppetserver-5.0.1.master-0.1SNAPSHOT.2017.08.22T1658.sles12.noarch.rpm
output/sles/12/PC1/i386/puppetserver-5.0.1.master-0.1SNAPSHOT.2017.08.22T1658.sles12.noarch.rpm
output/deb/jessie/PC1/puppetserver_5.0.1.master-0.1SNAPSHOT.2017.08.22T1658jessie_all.deb
output/deb/xenial/PC1/puppetserver_5.0.1.master-0.1SNAPSHOT.2017.08.22T1658xenial_all.deb
output/deb/stretch/PC1/puppetserver_5.0.1.master-0.1SNAPSHOT.2017.08.22T1658stretch_all.deb
```

melissa:
I think that, since this is currently fully backward compatible, and it's just in the early stages of getting this fully flushed out, we should move forward with it. I'm +1 to getting it merged, and for folks to start playing with it and giving us feedback.

haus:
How does this new workflow render failures in the rpm build or the deb build?

underscorgan:
hey @MosesMendoza, I'm going to try to address your questions but if something isn't making sense maybe it would be better to sync up in question? I feel like some of this is sort of complicated to explain clearly asynchronously :)

> Just to make sure I understand correctly, the new container-based build hosts will require the build tools installed already? I.e., we'll make (or already have) a fleet of container-based builders?

I've built some docker images (one for foss, one for PE) loaded with the requirements to use FPM to build packages. These images can be loaded into mesos with a label and then the packaging jobs will be bound to running on boxes with that label. 

> I'm not connecting the dots from this PR to enabling that future. Why does ezbake as-is require that we have static rpm-builder and deb-builder machines? This seems like something dictated by the packaging repo, but I could (definitely) have old information. Could we just update the job templates in packaging repo so that they could use ABS / check out the container-based hosts instead of using the static builders? We could also update these same job templates to instead use the new form of triggering existing jenkins jobs with data rather than creating the jobs dynamically? It seems like that would be of use to all projects (not just ezbake projects) and then we wouldn't have to do much to ezbake itself? I think I'm missing something.

It is packaging that requires those builder machines. Early in this process I evaluated making the changes to packaging instead of ezbake, but that was going to be a more complicated process so we opted to not do that at this time. Basically it comes down to `uber_build` takes packaging artifacts (spec files, control/rules/etc), gets them onto the builder boxes with the build environments/mocks/cows/etc, and produces packages. FPM, on the other hand, doesn't take in the same packaging artifacts, but rather has a command generated that encapsulates all the information from the spec or control files. There wasn't an obvious/straight-forward way to generalize this conversion in a way that would work across all the projects, so I decided the best way forward was to do this conversion for ezbake since it would have the most bang for our buck, and the few other projects still using straight packaging repo for packaging can go through a similar conversion (or conversion to vanagon if there's compiled code) in the future.

These containerized build machines aren't really a drop-in replacement for the existing builders. In the new world we won't be building the packages on os-specific machines. This is possible for ezbake since this is all java, and ezbake packages essentially make a/some jar(s) and configuration files and put them into an rpm/deb. This also means that we don't need to worry about building cows and mocks to enable new platforms, we can just start building packages for these new platforms (unless there are config/script/init-system changes needed). So, essentially we're redoing everything about how ezbake generates packages, but in a transparent manner. No changes needed in the projects for input, and the output is in the same format on builds.

> Separately, in terms of "future backwards-compatibility", we've run into issues in the past with packaging system changes that made it difficult to checkout older versions of our projects (ie in the case of security incidents). Do you know if we're going to need to make changes in the projects themselves to accommodate this future? (that's usually where it gets sticky is trying to backport these changes under duress later)

The only change I've needed to make to projects is updating the ezbake version.

This response got a little long, hope it helps clarify some things!

underscorgan:
@haus the errors will be whatever is produced by FPM.  I'll try to make sure we have enough output to make it useful (at least what platform was being packaged when we encountered the failure) and that a failure causes a job failure.

PR #433, opened 2017-08-10
(SERVER-249) Conditionally create directory for user JARs

Magisus:
This commit creates a specific location for users to put arbitrary JARs
that they want loaded on the classpath when Puppetserver starts. It will
only be created if `create-jar-dir` is set to true in the project's
EZBake config.

### DIFF
--- a/resources/puppetlabs/lein-ezbake/staging-templates/ezbake.rb.mustache
+++ b/resources/puppetlabs/lein-ezbake/staging-templates/ezbake.rb.mustache
@@ -52,7 +52,8 @@ module EZBake
                             "{{package}}" => "{{version}}",
                           {{/replaces-pkgs}}
                           },
-      :bootstrap_source => '{{{bootstrap-source}}}',
-      :logrotate_enabled => {{{logrotate-enabled}}}
+      :bootstrap_source  => '{{{bootstrap-source}}}',
+      :logrotate_enabled => {{{logrotate-enabled}}},
+      :create_jar_dir    => {{{create-jar-dir}}}
   }
 end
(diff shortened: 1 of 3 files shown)


haus:
You still need to thread this setting through the core and into ezbake.rb.mustache for this to work correctly.

Magisus:
Ooh. Can you give me a few more details on what is needed? Never worked in this project before.

haus:
@Magisus coming right up

haus:
Settings get added here: https://github.com/puppetlabs/ezbake/blob/master/src/puppetlabs/ezbake/core.clj#L464-L527
One they are there, you can add them to this mustache file: https://github.com/puppetlabs/ezbake/blob/master/resources/puppetlabs/lein-ezbake/staging-templates/ezbake.rb.mustache and the packaging loads that so it will be available to use.

haus:
@Magisus if you need more details than that I can probably fill them in as well

Magisus:
Thanks, I'll take a look!

Magisus:
@haus updated based on logicking about surrounding examples, let me know if I confused anything.

haus:
@Magisus so i forgot you're also going to want to make sure the packages actually own the dirs, which will involve diving into the spec file here: https://github.com/puppetlabs/ezbake/blob/master/resources/puppetlabs/lein-ezbake/template/global/ext/redhat/ezbake.spec.erb#L284

It does sort of look like you could just append the create_jar_dir to the list of dirs in create_dirs maybe?

Magisus:
So is that something I would want to set in Puppetserver's project.clj, right? I don't see that option in there yet, which could mean we don't need any special ones (until now).

Magisus:
On second thought, do I even need the conditional? Maybe all I need to make this happen is to use that `create_dirs` setting in Puppetserver's `project.clj`. That would mean I won't need any changes to EZBake at all.

haus:
@Magisus yea using create_dirs seems like it should be fine here. And it looks like it's designed to take a fully qualified path.

Magisus:
Cool, gonna close this PR then. Just did a build using create-dirs and verified that it created a directory.

PR #434, opened 2017-08-15
(EZ-114) Make startup port detection less naive

binford2k:


resources/puppetlabs/lein-ezbake/template/global/ext/ezbake-functions.sh.erb
```diff
@@ -21,7 +21,8 @@ wait_for_app()
         fi
 
         # if there are any TCP ports associated with the process, return success
-        netstat -tulpn 2>/dev/null | grep "$pid" 2>&1 >/dev/null
+        # ports are listed in column 7 for tcp and column 6 for udp. We don't care about udp.
```
shrug: Is this consistent across the various supported platforms?

haus:
I don't believe `wait_for_app` is used at all in this version of ezbake. This version of ezbake keys off of trapperkeeper updating a file to indicate that the service has fully started. You'll probably want to target the 0.2.x branch.

binford2k:
That would explain why I couldn't find it being used in the startup script I was troubleshooting. This is likely no longer relevant then. Thanks!

PR #438, opened 2017-09-09
(RE-8989) Updates for ezbake builds

underscorgan:


underscorgan:
This should block on https://github.com/puppetlabs/ci-job-configs/pull/3415 so the jobs actually exist, but this is otherwise ready for review.

resources/puppetlabs/lein-ezbake/template/global/ext/fpm.rb
```diff
@@ -135,10 +135,10 @@
 options.dist = "#{options.operating_system}#{options.os_version}" if options.dist.nil?
 
 if options.debug
-  STDERR.puts "=========================="
-  STDERR.puts "OPTIONS HASH"
-  STDERR.puts options
-  STDERR.puts "=========================="
+  puts "=========================="
```
shrug: Why change debug output from stderr to stdout?

underscorgan: I originally had it in stderr because I was having issue with puts statements not making it to stdout but that seems to be working. Is stderr better for debug output? I always forget what's standard

shrug: Probably fine as long as it requires enabling the debug output

underscorgan: Yeah, debug is disabled by default

resources/puppetlabs/lein-ezbake/template/global/controller.sh.erb
```diff
@@ -13,7 +15,9 @@ build_os=$1
 build_ver=$2
 
 BASEPATH='pkg_artifacts'
-mkdir $BASEPATH
+if [ ! -d "$BASEPATH" ]; then
+  mkdir $BASEPATH
+fi
```
mwaggett: Why not `mkdir -p`?

resources/puppetlabs/lein-ezbake/template/global/tasks/build.rake
```diff
@@ -32,14 +48,82 @@ namespace :pl do
...
         platform = cow.split('-')[1..-2].join('-')
-        platform_path = "deb/#{platform}/#{repo}"
+
+        # Keep on keepin' on with hardcoded paths in packaging
+        # Hopefully this goes away with packaging 1.0.x.
+        #  --MMR, 2017-08-30
+        if Pkg::Config.build_pe
+          platform_path = "pe/deb/#{platform}"
+        else
+          platform_path = "deb/#{platform}/#{repo}"
```
mwaggett: Will `platform` still have dashes here? Don't we want slashes by this point?

mwaggett: oh wait, don't we want the codename?

underscorgan: With all of the cows we have `platform` is set to the codename
`'base-jessie-i386.cow'split('-')[1..-2].join('-')` = `'jessie'`.  The join should never actually end up inserting a '-' unless the format of the cows change, it's mostly there to convert the array to a string and leave it set to something reasonable that we'd be more likely able to debug in case the format changes

resources/puppetlabs/lein-ezbake/template/global/controller.sh.erb
underscorgan: This felt more explicit for what I was trying to do. I've mostly used `mkdir -p` when making a directory down a few levels, so running it when making a directory in PWD feels strange :woman_shrugging:

resources/puppetlabs/lein-ezbake/template/global/tasks/build.rake
```diff
@@ -32,14 +48,82 @@ namespace :pl do
...
+      end
+      Pkg::Util::Net.print_url_info(args[:job_url])
+    end
+    desc "trigger jenkins packaging job with local auth"
+    task :trigger_build_local_auth => "pl:fetch" do
+      if Pkg::Config.build_pe
+        job_url = "https://cinext-jenkinsmaster-enterprise-prod-1.delivery.puppetlabs.net/job/enterprise_various-packaging-jobs_packaging-os-clj_lein-ezbake-generic"
+      else
+        job_url = "https://jenkins-master-prod-1.delivery.puppetlabs.net/job/platform_various-packaging-jobs_packaging-os-clj_lein-ezbake-generic"
+      end
```
mwaggett: Is there a way to not hard code this? Or at least DRY it up:
```
if Pkg::Config.build_pe
  jenkins = "cinext-jenkinsmaster-enterprise-prod-1"
  stream = "enterprise"
else
  jenkins = "jenkins-master-prod-1"
  stream = "platform"
end
job_url = "https://#{jenkins}.delivery.puppetlabs.net/job/#{stream}_various-packaging-jobs_packaging-os-clj_lein-ezbake-generic"
```

resources/puppetlabs/lein-ezbake/template/global/tasks/build.rake
mwaggett: ooh, nice. yeah, it was the `join` that made me assume `plat-ver`. 👍

resources/puppetlabs/lein-ezbake/template/global/controller.sh.erb
mwaggett: hmm. I see what you mean, though I can't claim to know what's idiomatic. I'm just generally a fan of shorter/cleaner. 
also, theoretically, couldn't `$BASEPATH` have several levels?

underscorgan: `BASEPATH` is hard-coded one line above this, so it will always have one level.

resources/puppetlabs/lein-ezbake/template/global/tasks/build.rake
underscorgan: These need to be hardcoded to maintain current functionality for kicking off a build from the command line with an `ezbake build`. I can make the suggested change though.

resources/puppetlabs/lein-ezbake/template/global/controller.sh.erb
mwaggett: oh yeah. It could theoretically change though..? idk, I don't have strong feelings. ¯\_(ツ)_/¯

underscorgan: if we change it we can change how the directory is made though, since we'll be updating this file anyways. In bash I tend to prefer verbosity if it makes it more clear/explicit, which I think this does.

mwaggett: fair enough 😊

resources/puppetlabs/lein-ezbake/template/global/tasks/build.rake
```diff
@@ -11,19 +11,35 @@ namespace :pl do
...
+        puts "===================================="
+        puts "Packaging for #{os} #{ver}"
+        puts "===================================="
+        stdout, stderr, exitstatus = Pkg::Util::Execution.capture3(%(bash controller.sh #{os} #{ver}))
+        Pkg::Util::Execution.success?(exitstatus) or raise "Error running packaging: #{stdout}\n#{stderr}"
+        puts "#{stdout}\n#{stderr}"
+
+        # I'm so sorry
+        # These paths are hard-coded in packaging, so hard code here too.
+        # When everything is moved to packaging 1.0.x this should be able
```
melissa: It looks like even moving to 1.0.x we don't get to fix this :( However, moving to artifactory, we will!

src/puppetlabs/ezbake/core.clj
```diff
@@ -819,6 +819,14 @@ Additional uberjar dependencies:
   (action "stage" lein-project build-target)
   (exec/exec "rake" "package:bootstrap" :dir staging-dir)
   (let [downstream-job nil
+        rake-call ["rake" "pl:jenkins:trigger_build_local_auth"]]
```
melissa: So if I call `lein with-profile ezbake ezbake build`, that will invoke this new way of building? How do the pipelines kick off the ezbake build? I'm wondering if this behavior change should trigger a new major version? If the existing ci pipelines build with a different call that they can opt into beyond just bumping the ezbake version, that'd be fantastic.

Also, we should update the docs. I know I refer to https://confluence.puppetlabs.com/display/ENG/How+To+build+Puppet+Server+from+a+local+topic+branch a lot when building locally with ezbake changes

resources/puppetlabs/lein-ezbake/template/global/tasks/build.rake
underscorgan: ok, cool. I can update the comment

src/puppetlabs/ezbake/core.clj
underscorgan: None of the pipelines use `lein with-profile ezbake ezbake build` but rather call `uber_ship` directly, so this would only affect testing locally, and the change should be completely transparent with the exception of needing to set the `JENKINS_USER_AUTH` environment variable to auth with jenkins. I tried to make this as straightforward a transition as possible.

underscorgan: The only change for running from an ezbake topic branch is setting `JENKINS_USER_AUTH`. Once it's merged/released I'll update confluence. It will also fail very informatively and let you know that this var is needed and how to find/set it.

resources/puppetlabs/lein-ezbake/template/global/tasks/build.rake
```diff
@@ -32,14 +48,86 @@ namespace :pl do
...
+      if args[:auth_string] =~ /:/
+        curl_opts << "--user #{args[:auth_string]}"
+        Pkg::Util::Net.curl_form_data("#{args[:job_url]}/build", curl_opts)
+      #assume we're using a token
+      else
+        Pkg::Util::Net.curl_form_data("#{args[:job_url]}/build?token=#{args[:auth_string]}", curl_opts)
+      end
+      Pkg::Util::Net.print_url_info(args[:job_url])
+    end
+    desc "trigger jenkins packaging job with local auth"
```
melissa: personally I prefer spaces between the different tasks, but that's not a blocker.

resources/puppetlabs/lein-ezbake/template/global/tasks/build.rake
```diff
@@ -32,14 +48,86 @@ namespace :pl do
...
+      curl_opts = [
+        '--request POST',
+        "--form file0=@#{Dir.pwd}/BUILD_PROPERTIES",
+        "--form file1=@#{Dir.pwd}/PROJECT_BUNDLE",
+      ]
+      if Pkg::Config.build_pe
+        Pkg::Util.check_var('PE_VER', ENV['PE_VER'])
+        curl_opts << %(--form json='{"parameter": [{"name":"PE_VER", "value":"#{ENV['PE_VER']}"},{"name":"BUILD_PROPERTIES", "file":"file0"},{"name":"PROJECT_BUNDLE", "file":"file1"}]}')
+      else
+        curl_opts << %(--form json='{"parameter": [{"name":"BUILD_PROPERTIES", "file":"file0"},{"name":"PROJECT_BUNDLE", "file":"file1"}]}')
```
melissa: Are these basically the same as what we have in the packaging repo?

resources/puppetlabs/lein-ezbake/template/global/tasks/build.rake
```diff
@@ -32,14 +48,86 @@ namespace :pl do
...
+      #assume we're using a token
+      else
+        Pkg::Util::Net.curl_form_data("#{args[:job_url]}/build?token=#{args[:auth_string]}", curl_opts)
+      end
+      Pkg::Util::Net.print_url_info(args[:job_url])
+    end
+    desc "trigger jenkins packaging job with local auth"
+    task :trigger_build_local_auth => "pl:fetch" do
+      if Pkg::Config.build_pe
+        jenkins = 'cinext-jenkinsmaster-enterprise-prod-1'
```
melissa: I want us to not hard code these values. Is there a way we could allow someone to run this outside of our infrastructure? That would require they set up an identical job to what we have on these two jenkins servers, but that would make us want to allow users to override these settings

melissa: I'm not completely blocking on this, given that the previous behavior was also locked in to our infrastructure. But I think it would be a good thing to think about.

melissa:
I think it's generally good. I'm curious about the overlap between the automation you're adding in and that in the packaging repo though. That is something we should make sure we do right, given how many times it's bitten us before. Other than that, the style nits are definitely not blockers to getting this merged.

resources/puppetlabs/lein-ezbake/template/global/tasks/build.rake
underscorgan: It's passing everything as a single json block here instead of as individual arguments, but doing basically the same thing.

resources/puppetlabs/lein-ezbake/template/global/tasks/build.rake
underscorgan: Yeah, I was thinking about this. I think we could add the full url to the jenkins job as a parameter in ezbake that we have sensible defaults for, but requiring that users then reverse-engineer our jenkins job to actually do anything with it didn't feel much better than it being hard-coded.

resources/puppetlabs/lein-ezbake/template/global/tasks/build.rake
melissa: Are we pulling in any functionality from the packaging repo for any of this?

underscorgan: Just the curl_form_data and shell execution utilities. This functionality in packaging didn't really feel generalized enough to just drop it in place.

underscorgan:
@melissa I tried to use packaging resources when they made sense and only add things when they were fairly specific to this workflow. I'm sure there are things that could be cleaned up more or combined into utilities, I didn't focus a lot of energy on that. I made sure none of the rake tasknames were overloaded so we at least know what's actually going to get called

melissa:
I merged because there were only minimal changes between now and when Rob approved.

PR #440, opened 2017-09-15
(maint) fixup fpm bugs

underscorgan:


resources/puppetlabs/lein-ezbake/template/global/controller.sh.erb
```diff
@@ -3,60 +3,61 @@
...
 # and setup needed to build packages for <os>
 # The os string should be one of el, sles, debian, ubuntu,
 # or fedora
 # The version should be the numerical OS version for el,
 # sles, and fedora and the codename for debian/ubuntu.
+# The resulting artifacts will be stored in <basepath>
 ################
 build_os=$1
 build_ver=$2
+basepath=$3
```
melissa: Should we have a default so that folks aren't required to set basepath?

demophoon: oops, i merged already :X

underscorgan: :woman_shrugging: I don't know that this script is ever going to be used manually, it just gets executed from rake tasks

demophoon: That was my understanding of this script

underscorgan: can always fix it later if that usecase ends up happening

PR #444, opened 2017-09-25
(maint) Make sure the packaging tarball is shipped

underscorgan:
Initially we were generating the packaging tarball then moving it into a
staging directory to prepare artifacts for the build step. However, we
should be shipping this tarball to builds, so let's copy it instead.

### DIFF
--- a/resources/puppetlabs/lein-ezbake/template/global/tasks/build.rake
+++ b/resources/puppetlabs/lein-ezbake/template/global/tasks/build.rake
@@ -10,7 +10,7 @@ namespace :pl do
     nested_output = '../../../output'
     pkg_path = '../pkg'
     staging_path = 'pkg_artifacts'
-    FileUtils.mv(Dir.glob("pkg/*.gz").join(''), FileUtils.pwd)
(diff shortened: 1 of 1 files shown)


underscorgan:
Verified the tarball is being shipped with http://builds.delivery.puppetlabs.net/pe-console-services/1.9.213.SNAPSHOT.2017.09.25T1209/artifacts/

resources/puppetlabs/lein-ezbake/template/global/tasks/build.rake
```diff
@@ -10,7 +10,7 @@ namespace :pl do
     nested_output = '../../../output'
     pkg_path = '../pkg'
     staging_path = 'pkg_artifacts'
-    FileUtils.mv(Dir.glob("pkg/*.gz").join(''), FileUtils.pwd)
+    FileUtils.cp(Dir.glob("pkg/*.gz").join(''), FileUtils.pwd)
```
e-gris: <Include standard not-helpful, gratuitous, anti-glob rant here>

PR #450, opened 2017-10-30
(maint) Update EZBake to pull from the 1.0.x branch of packaging

melissa:
This will enable us to utilize new and improved utilities that are
available in this branch.


melissa:
I built both puppetserver and puppetdb 5.x branches with this. We shouldn't even attempt to update those projects with this ezbake until after the releases are out next week

PR #452, opened 2017-11-14
(maint) We accidentally released 1.7.0 as 1.6.4.  Let's fix this up good.

underscorgan:
Especially since I found a bug that's going to require a 1.6.5, let's revert all the 1.7.0 stuff until that gets in, then rebuild 1.7.0 with the triggers stuff.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,24 +1,3 @@
-## 1.7.0 - 2017-11-08
-This is a feature and maintenance release.
-
-Feature:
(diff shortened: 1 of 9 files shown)


underscorgan:
@mwaggett no, the params don't do what we think they do. The version that gets released trims off the -SNAPSHOT in the version string, and the parameter determines what the next -SNAPSHOT release is, so it tagged it as 1.6.4 and then made the next snapshot 1.7.0-SNAPSHOT

PR #458, opened 2017-11-17
(maint) Pass MOCK and COW settings when running `lein ezbake build`

underscorgan:
The generic ezbake jenkins jobs now have MOCKS and COWS parameters to
help with determining repo creation. We should make sure we're passing
local mock/cow values through so we create repos correctly.


underscorgan:
cherry-picked from master

PR #461, opened 2017-11-29
(maint) Undo some bad assumptions from packaging hard-coding

underscorgan:
For debian repos with no apt_repo_name set, the packaging tasks that
generate the repo look for artifacts under deb/dist, rather than
deb/dist/main as it had originally appeared. This fixes the code to
match the assumptions in packaging.

### DIFF
--- a/resources/puppetlabs/lein-ezbake/template/global/tasks/build.rake
+++ b/resources/puppetlabs/lein-ezbake/template/global/tasks/build.rake
@@ -48,17 +48,21 @@ namespace :pl do
         end
       end
       Pkg::Config.cows.split(" ").each do |cow|
-        # carry forward defaults from Packaging::Deb::Repo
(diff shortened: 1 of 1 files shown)


resources/puppetlabs/lein-ezbake/template/global/tasks/build.rake
```diff
@@ -58,7 +62,8 @@ namespace :pl do
         if Pkg::Config.build_pe
```
e-gris: My standard gripe. I would rather see it this way to create a sense of normal/exceptional case.

```ruby
platform_path = "pe/deb/#{platform}"
if !Pkg::Config.build_pe
  platform_path = "deb/#{platform}/#{repo}".sub(/\/$/, '')
end
```

underscorgan:
@e-gris updated!

PR #469, opened 2017-12-18
(SERVER-2068) Be more discerning finding PIDs

reidmv:
Previously, the start/stop/reload scripts would use a `pgrep` to search
for running processes, to determine if the service was still running.
That `pgrep` would do a simple search for uberjar name.

Turns out, that's not specific enough. In the case of puppetserver, a
user invocation of `puppetserver gem list` will put a process in the
process list that the naive `pgrep` would finger as the service pid in
the event puppetserver wasn't actually running. This could trick a
service manager into thinking puppetserver was running when in fact it
could be stopped for hours.

This commit doesn't change the basic service detection logic, but it
improves the `pgrep` to be more discerning, and search not only for the
uberjar name but also for the main namespace.

### DIFF
--- a/resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
+++ b/resources/puppetlabs/lein-ezbake/template/global/ext/cli/reload.erb
@@ -17,7 +17,7 @@ fi
 init_restart_file "$restartfile" || exit $?
 
 initial="$(head -n 1 "$restartfile")"
-pid="$(pgrep -f <%= EZBake::Config[:uberjar_name] %>)"
+pid="$(pgrep -f "<%= EZBake::Config[:uberjar_name] %>.* -m <%= EZBake::Config[:main_namespace] %>")"
 kill -HUP $pid >/dev/null 2>&1
 if [ $? -ne 0 ]; then
     echo "Service not running so cannot be reloaded" 1>&2
--- a/resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
+++ b/resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set +e
 
-pid="$(pgrep -f <%= EZBake::Config[:uberjar_name] %>)"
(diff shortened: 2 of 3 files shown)


reidmv:
@shrug I'm not familiar with this code, but I'm assuming the namespace for the running service is known to be what we configure in the start script: https://github.com/puppetlabs/ezbake/pull/469/files#diff-c86ad2f37a924af86083d5a858b2de69R66.

In the ticket, I have [examples](https://tickets.puppetlabs.com/browse/SERVER-2068?focusedCommentId=515693&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-515693) of the specific problem I ran into. They showed a difference between the service invocation and a CLI invocation as being e.g.

```
clojure.main -m puppetlabs.trapperkeeper.main
clojure.main -m puppetlabs.puppetserver.cli.gem
```

reidmv:
One thing this won't account for with puppetserver specifically, now that I'm thinking about it. The `puppetserver foreground` command will still get picked up by this `pgrep` as the service.

This is still better because `puppetserver foreground` conflicts with the actual service and so is way less likely to cause confusion or be a problem during normal operation. But, the scenario shows that a deeper logic fix would be required to fill all the cracks completely.

This PR is still an improvement and will better the user experience in 90% of use cases. I still advocate incorporating it as an iterative improvement, until or unless we prioritize a more complete fix.

reidmv:
@shrug another logic option that'd work on EL 7 would be to modify pgrep with the -P option (--parent). E.g. `pgrep --parent 1 puppetlabs.trapperkeeper.main`. I don't know if PPID=1 is consistent across all init systems though.

mwaggett:
@reidmv Were you able to test this out and get the result you were looking for?

reidmv:
@mwaggett waiting to test but should have some feedback soon. We know the more discerning pgrep works at least as well as the old one, the open question still is whether or not the extra specificity is enough to solve the problem.

Keep in mind I don't see this PR as a perfect solution even if it works for the puppetserver use case. It's reinforcing and firming up logic that will still have an underlying weakness. Worth merging even so, if it works? Yes, because it's low-risk and high-reward.

I'll report back after I get some information on whether or not it fixed the issue for the customer in production.

mwaggett:
@reidmv Sounds good, yeah, happy to merge even if it's just a small improvement.

PR #472, opened 2018-02-07
(CPR-515) Allow separate tasks for install vs upgrade

shrug:
Currently, the postinstall tasks from the packaged project
are executed unconditionally, both on install as well as
upgrades. This can cause issues, as in the case with puppetserver
potentially overwriting config values if the user has changed them
from the package defaults.

This adds an addtional set of tasks that can be executed in the
package post-install script, that are only executed on initial
package install. This leaves the current behavior untouched, so
packages depending on the current behavior can be unmodified.


shrug:
We could add another section for upgrades, add them to other scriptlets, etc. This is the minimum needed to satisfy the current requirement.

shrug:
Another way forward would be to wrap each task in, for example, https://github.com/shrug/puppet-server/blob/a523c1ca4ab9924fc8f886ece30cf193e4be6d7c/resources/ext/ezbake.conf#L24-L34 in conditionals depending on the requirements. That seems less maintainable, but we already do that in a limited fashion in PuppetDB, for example.

resources/puppetlabs/lein-ezbake/template/global/install.sh.erb
```diff
@@ -327,22 +327,38 @@ function task_preinst_deb {
...
 <% EZBake::Config[:debian][:additional_postinst].each do |cmd| -%>
     <%= cmd %>
 <% end -%>
 }
 
-# RPM based post-installation tasks.
+# Debian based install post-installation tasks.
+function task_postinst_deb_install {
+    task postinst_permissions
+<% EZBake::Config[:debian][:additional_postinst_install].each do |cmd| -%>
```
mwaggett: Do you need a null command for this one too?

resources/puppetlabs/lein-ezbake/template/global/ext/debian/postinst.erb
```diff
@@ -2,6 +2,11 @@
 
 /opt/puppetlabs/server/apps/<%= EZBake::Config[:real_name] -%>/scripts/install.sh postinst_deb
 
+# Set some configuration from the project on install
+if [ -z "$2" ]; then
```
mwaggett: Where's the value of `$2` coming from?
also, is it meant to be in quotes?

resources/puppetlabs/lein-ezbake/template/global/ext/redhat/ezbake.spec.erb
```diff
@@ -223,7 +223,13 @@ fi
 <% end -%>
 
 %post
+# Unconditional post-install tasks
 %{_app_prefix}/scripts/install.sh postinst_redhat
+
+# Only run these on initial install, not ugrades
+if [ "$1" = "1" ]; then
```
mwaggett: same questions about values and quotes

resources/puppetlabs/lein-ezbake/template/global/ext/debian/postinst.erb
underscorgan: `$2` is a variable available in debian packaging, set to the previous version if this is an upgrade, not set if it's a fresh install. Yes, it needs to be in quotes, we're testing to see whether or not this is an empty string

resources/puppetlabs/lein-ezbake/template/global/ext/redhat/ezbake.spec.erb
underscorgan: redhat does this a little differently, but still a variable that lets you know whether this is an install/upgrade/etc. See the chart at https://fedoraproject.org/wiki/Packaging:Scriptlets. You need to quote it for the bash comparison

resources/puppetlabs/lein-ezbake/template/global/install.sh.erb
underscorgan: Nope, since there's the task postinst_permissions in this function already. That's only needed if you're going to end up with an empty block

underscorgan:
Assuming this works as expected on test, 👍

shrug:
Test build: http://builds.delivery.puppetlabs.net/puppetserver/5.1.6.SNAPSHOT.2018.02.07T1415/

resources/puppetlabs/lein-ezbake/template/global/install.sh.erb
```diff
@@ -327,22 +327,38 @@ function task_preinst_deb {
...
     task postinst_permissions
 <% EZBake::Config[:debian][:additional_postinst].each do |cmd| -%>
     <%= cmd %>
 <% end -%>
 }
 
-# RPM based post-installation tasks.
+# Debian based install post-installation tasks.
+function task_postinst_deb_install {
+    task postinst_permissions
```
haus: Will calling postinst_permissions twice cause problems?

shrug: oh yeah we should drop that. copy-pasta

resources/puppetlabs/lein-ezbake/template/global/install.sh.erb
```diff
@@ -327,22 +327,37 @@ function task_preinst_deb {
...
     <%= cmd %>
 <% end -%>
 }
 
-# RPM based post-installation tasks.
+# Debian based install post-installation tasks.
+function task_postinst_deb_install {
+<% EZBake::Config[:debian][:additional_postinst_install].each do |cmd| -%>
+    <%= cmd %>
+<% end -%>
```
underscorgan: You need a null command in here now that you've removed the postinst_permissions

resources/puppetlabs/lein-ezbake/template/global/ext/redhat/ezbake.spec.erb
```diff
@@ -223,7 +223,13 @@ fi
...
+  %{_app_prefix}/scripts/install.sh postinst_redhat_install
+fi
```
underscorgan: With the FPM stuff you actually want to be updating this file, https://github.com/puppetlabs/ezbake/blob/master/resources/puppetlabs/lein-ezbake/template/global/ext/redhat/postinst.erb

The spec file is only used for legacy builds. I want to burn that down for exactly this reason Soon, but it's not there yet

sad

shrug: oh cool, ok I was wondering how the standalone files were being used.

shrug:
Updated to address comments

shrug:
This now works as expected for RPM - installs set the config variables, upgrades do not overwrite them.

shrug:
Also works on debian. 

Tested with http://builds.delivery.puppetlabs.net/puppetserver/5.1.6.SNAPSHOT.2018.02.07T1610/

PR #474, opened 2018-02-08
(RE-10173) Merge master -> stable

underscorgan:
Merge master back to stable to pick up the build improvements, features, and updates required to ship to artifactory along with builds.

Maintains java dependencies from stable, and adds back support for wheezy.


underscorgan:
Also, after this goes in I think we should probably push stable -> legacy and delete the stable branch. Stable really is a misnomer here, it's not 'stable', it just supports EOL platforms/javas.

PR #476, opened 2018-02-14
(maint) 1.7.6 prep

underscorgan:


CHANGELOG.md
```diff
@@ -1,3 +1,14 @@
+## 1.7.6 - 2018-02-14
+This is a bugfix release.
+
+Bugfix:
+  * Reverts commit setting `build_tar: FALSE` in foss build defaults. This fix
+    was masking a bug in github.com/puppetlabs/packaging which has since been
+    fixed.
+  * (CPR-515) It was not previously possible to specify that postinstall tasks
```
underscorgan: I was torn for this one. It's technically a new feature in ezbake, but was added for a bugfix. 🤷‍♀️

mwaggett: Does this change affect how people would be using ezbake? Like, will people have to change what param names they use in order to maintain old functionality? I forget the specifics of the change, but I assume not, so I'm cool calling this a bugfix.

sarameisburger: Nothing they need to change, only if they want something only to be run on install. Let's go with bugfix

underscorgan: You don't have to change anything to maintain existing functionality, but, say, to fix the issue reported against puppetserver you will need to change you ezbake config (whichever tasks should only run on install will need to be moved from the `postinst` section https://github.com/puppetlabs/puppetserver/blob/5.1.x/resources/ext/ezbake.conf#L23-L35 to a `postinst-install` section.

sarameisburger: Hmmm what's the definition of a feature? Added functionality?

underscorgan: from semver.org:
> Given a version number MAJOR.MINOR.PATCH, increment the:
>
> MAJOR version when you make incompatible API changes,
> MINOR version when you add functionality in a backwards-compatible manner, and
> PATCH version when you make backwards-compatible bug fixes.

mwaggett: starting to think this is a feature release then, but I don't have super strong feelings

sarameisburger: Just call it a bugture

project.clj
```diff
@@ -1,4 +1,4 @@
-(defproject puppetlabs/lein-ezbake "1.7.6-SNAPSHOT"
+(defproject puppetlabs/lein-ezbake "1.8.0-SNAPSHOT"
```
mwaggett: You'd want this as `1.8.1-SNAPSHOT` right? since it points to whatever is next?

CHANGELOG.md
```diff
@@ -1,3 +1,14 @@
+## 1.8.0 - 2018-02-14
+This is a feature and bugfix release.
+Feature:
```
sarameisburger: `Feature:` isn't showing up on a new line

project.clj
underscorgan: Nope, the jenkins job will do that. I just need to get it to 1.8.0-SNAPSHOT so it doesn't publish this as 1.7.6

CHANGELOG.md
underscorgan: fixed!

PR #478, opened 2018-02-17
(maint) A few bugfixes found while testing a puppetserver update

underscorgan:
These commits should be cherry-picked up to master


mwaggett:
How did you test this?

underscorgan:
I've been running builds and then comparing package contents

mwaggett:
How were we getting by with these bugs in the first place?

Everything looks reasonable, just trying to understand things more.

underscorgan:
They're not bugs that prevent packages from installing or services from running, they're just bugs that the fpm stuff introduced. I was doing a lot of testing with the initial conversion but there were testing and refactoring cycles, and clearly some things were missed.

Right now I'm comparing scripts, package requires, and files contained in packages built for the last puppetserver 2.x tagged build and then just updating to an ezbake snapshot built off of this branch

resources/puppetlabs/lein-ezbake/template/global/controller.sh.erb
```diff
@@ -105,10 +105,27 @@ params=("--user" "<%= EZBake::Config[:user] -%>" "--group" "<%= EZBake::Config[:
...
+<% EZBake::Config[:terminus_info].each do |proj_name, info| -%>
+<% if comma_counter == 0
+     comma = ', '
+   else
+     comma = ''
+   end
+   comma_counter += 1
+%>
+termini_description+='<%= comma -%><%= proj_name -%> (version <%= info[:version] -%>)'
+<% end -%>
```
underscorgan: The old packaging used newlines in its' descriptions, but I was having a hard time getting that working, so I added comma logic. https://github.com/jordansissel/fpm/issues/1468

mwaggett: 1. How does erb feel about indentation? This took me a minute to parse out, so I feel like it could be helpful for human eyes if the `if` block and `termini_description+=` were indented from the loop.

2. I assume you tested and saw this work how you wanted, but I'm having trouble seeing how it works. Should maybe the values be switched? Like you wouldn't want a comma after the `for: ` i.e. when comma_counter is 0, but you would want it for the rest? And then you'd probably not want it for the last one again, so maybe a case where there's no comma if `comma_counter == EZBake::Config[:terminus_info].count`?

underscorgan: this is still being actively worked on, so yes, those were backwards.

I can probably get the if block better intented, but the termini_descriptor line is actually ending up in the shell script, so indenting it isn't going to work

mwaggett: ooh, sorry, didn't know you were still working on all this, I'll hold off future review til you're ready. 😊

resources/puppetlabs/lein-ezbake/template/global/controller.sh.erb
```diff
@@ -105,10 +105,27 @@ params=("--user" "<%= EZBake::Config[:user] -%>" "--group" "<%= EZBake::Config[:
...
+<% EZBake::Config[:terminus_info].each do |proj_name, info| -%>
+<%   if comma_counter == 0
+       comma = ''
+     else
+       comma = ', '
+     end
+     comma_counter += 1
+-%>
+termini_description+='<%= comma -%><%= proj_name -%> (version <%= info[:version] -%>)'
+<%   end # if comma_counter == 0 -%>
```
mwaggett: probably want this comment gone
also, this `end` goes with the `each`, right? so I think should be left aligned.

resources/puppetlabs/lein-ezbake/template/global/ext/fpm.rb
```diff
@@ -6,7 +6,6 @@
 
 options = OpenStruct.new
 # settin' some defaults
-options.certs_package = 'ca-certificates'
```
shrug: Do we not want that?

underscorgan: It looks like in ye olde packages that was just a build dependency that got mis-translated as a runtime dependency https://github.com/puppetlabs/ezbake/blob/master/resources/puppetlabs/lein-ezbake/template/global/ext/redhat/ezbake.spec.erb#L114-L117

PR #479, opened 2018-02-28
(EZ-20) EZBake defaults templates should create a separate variable f…

sarameisburger:
…or TrapperKeeper arguments

Adding TrapperKeeper-specific arguments to java_args, such as --debug, causes java to give up and exit.
This makes a separate variable, TK_ARGS, in the defaults file that could be used to adjust the flags passed to daemons and command line tools.

### DIFF
--- a/resources/puppetlabs/lein-ezbake/template/foss/ext/default.erb
+++ b/resources/puppetlabs/lein-ezbake/template/foss/ext/default.erb
@@ -8,6 +8,9 @@ JAVA_BIN="/usr/bin/java"
 # Modify this if you'd like to change the memory allocation, enable JMX, etc
 JAVA_ARGS="<%= EZBake::Config[:java_args] %>"
 
+# Modify this if you'd like TrapperKeeper specific arguments
+TK_ARGS="<%= EZBake::Config[:tk_args] %>"
+
 # These normally shouldn't need to be edited if using OS packages
 USER="<%= EZBake::Config[:user] %>"
 GROUP="<%= EZBake::Config[:group] %>"
--- a/resources/puppetlabs/lein-ezbake/template/global/ext/cli/foreground.erb
+++ b/resources/puppetlabs/lein-ezbake/template/global/ext/cli/foreground.erb
@@ -33,6 +33,7 @@ COMMAND="${JAVA_BIN} ${JAVA_ARGS} ${LOG_APPENDER} \
          clojure.main -m <%= EZBake::Config[:main_namespace] %> \
          --config ${CONFIG} --bootstrap-config ${BOOTSTRAP_CONFIG} \
          --restart-file "${restartfile}" \
+         ${TK_ARGS} \
          ${@}"
 
 pushd "${INSTALL_DIR}" &> /dev/null
--- a/resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
+++ b/resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
@@ -66,7 +66,8 @@ ${JAVA_BIN} ${JAVA_ARGS} -Djava.security.egd=/dev/urandom \
   -m <%= EZBake::Config[:main_namespace] %> \
   --config "${CONFIG}" \
   --bootstrap-config "${BOOTSTRAP_CONFIG}" \
-  --restart-file "${restartfile}" &
+  --restart-file "${restartfile}" \
+  ${TK_ARGS} &
 
 # $! is the process id of the last backgrounded process, the Java process above.
 pid=$!
--- a/resources/puppetlabs/lein-ezbake/template/pe/ext/default.erb
+++ b/resources/puppetlabs/lein-ezbake/template/pe/ext/default.erb
@@ -8,6 +8,9 @@ JAVA_BIN="/opt/puppetlabs/server/bin/java"
 # Modify this if you'd like to change the memory allocation, enable JMX, etc
 JAVA_ARGS="<%= EZBake::Config[:java_args] %>"
 
+# Modify this if you'd like TrapperKeeper specific arguments
+TK_ARGS="<%= EZBake::Config[:tk_args] %>"
+
 # These normally shouldn't need to be edited if using OS packages
 USER="<%= EZBake::Config[:user] %>"
 GROUP="<%= EZBake::Config[:group] %>"
--- a/src/puppetlabs/ezbake/core.clj
+++ b/src/puppetlabs/ezbake/core.clj
@@ -79,6 +79,7 @@
    (schema/optional-key :open-file-limit) schema/Int
    (schema/optional-key :main-namespace) schema/Str
    (schema/optional-key :java-args) schema/Str
+   (schema/optional-key :tk-args) schema/Str
    (schema/optional-key :redhat-postinst-install-triggers) RPMTriggers
    (schema/optional-key :redhat-postinst-upgrade-triggers) RPMTriggers
    (schema/optional-key :debian-interested-install-triggers) DEBTriggers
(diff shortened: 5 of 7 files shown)


resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
```diff
@@ -66,7 +66,8 @@ ${JAVA_BIN} ${JAVA_ARGS} -Djava.security.egd=/dev/urandom \
   -m <%= EZBake::Config[:main_namespace] %> \
   --config "${CONFIG}" \
   --bootstrap-config "${BOOTSTRAP_CONFIG}" \
-  --restart-file "${restartfile}" &
+  --restart-file "${restartfile}" \
+  ${TK_ARGS} &
```
sarameisburger: I'm not sure if this is where it should be added...

sarameisburger:
tested w/ puppetserver -- etc/sysconfig/puppetserver
```
###########################################
# Init settings for puppetserver
###########################################

# Location of your Java binary (version 7 or higher)
JAVA_BIN="/usr/bin/java"

# Modify this if you'd like to change the memory allocation, enable JMX, etc
JAVA_ARGS="-Xms2g -Xmx2g -Djruby.logger.class=com.puppetlabs.jruby_utils.jruby.Slf4jLogger"

# Modify this if you'd like TrapperKeeper specific arguments
TK_ARGS="-test"

# These normally shouldn't need to be edited if using OS packages
USER="puppet"
GROUP="puppet"
INSTALL_DIR="/opt/puppetlabs/server/apps/puppetserver"
CONFIG="/etc/puppetlabs/puppetserver/conf.d"

# Bootstrap path
BOOTSTRAP_CONFIG="/etc/puppetlabs/puppetserver/services.d/,/opt/puppetlabs/server/apps/puppetserver/config/services.d/"

# SERVICE_STOP_RETRIES can be set here to alter the default stop timeout in
# seconds.  For systemd, the shorter of this setting or 'TimeoutStopSec' in
# the systemd.service definition will effectively be the timeout which is used.
SERVICE_STOP_RETRIES=60

# START_TIMEOUT can be set here to alter the default startup timeout in
# seconds.  For systemd, the shorter of this setting or 'TimeoutStartSec'
# in the service's systemd.service configuration file will effectively be the
# timeout which is used.
START_TIMEOUT=300


# Maximum number of seconds that can expire for a service reload attempt before
# the result of the attempt is interpreted as a failure.
RELOAD_TIMEOUT=120
```

src/puppetlabs/ezbake/core.clj
```diff
@@ -568,6 +569,8 @@ Additional uberjar dependencies:
                                                       "puppetlabs.trapperkeeper.main")
      :java-args                          (get-local-ezbake-var lein-project :java-args
                                                       "-Xmx192m")
+     :tk-args                            (get-local-ezbake-var lein-project :tk-args
+                                                      "-test")
```
mwaggett: What's this `"-test"` bit for?

sarameisburger: whoops forgot to remove that, it was for testing

src/puppetlabs/ezbake/core.clj
```diff
@@ -568,6 +569,8 @@ Additional uberjar dependencies:
                                                       "puppetlabs.trapperkeeper.main")
      :java-args                          (get-local-ezbake-var lein-project :java-args
                                                       "-Xmx192m")
+     :tk-args                            (get-local-ezbake-var lein-project :tk-args
+                                                       "" )
```
sarameisburger: @mwaggett updated. I had to set it to an empty string otherwise it gets mad
```
Error encountered performing task 'ezbake' with profile(s): 'ezbake'
clojure.lang.ArityException: Wrong number of args (2) passed to: core/get-local-ezbake-var
	at clojure.lang.AFn.throwArity(AFn.java:429)
	at clojure.lang.AFn.invoke(AFn.java:36)
	at puppetlabs.ezbake.core$make_template_map.invokeStatic(core.clj:572)
```

mwaggett: I'd like to know what that value is used for. Is it a default?

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
mwaggett: ^ same tbh. I should up my ezbake knowledge.

src/puppetlabs/ezbake/core.clj
sarameisburger: Yes, it's used when creating the ezbake.rb template. I'm waiting on the ticket creator to get what the default value should be.

mwaggett:
@sarameisburger Would it be possible to see the bad behavior, before this change, and then the fixed behavior with it? like with a real trapperkeeper argument?

resources/puppetlabs/lein-ezbake/template/global/ext/cli/start.erb
Sharpie: This looks good to me. Should probably be copied over to the `foreground` command as well, before the `${@}` expansion. The `stop` and `reload` commands probably don't need the extra arguments.

src/puppetlabs/ezbake/core.clj
Sharpie: Empty string is a good default. Most common use of this would be to get the `--debug` flag passed into TK daemon(s) when needed.

resources/puppetlabs/lein-ezbake/template/global/ext/cli/foreground.erb
```diff
@@ -33,6 +33,7 @@ COMMAND="${JAVA_BIN} ${JAVA_ARGS} ${LOG_APPENDER} \
          clojure.main -m <%= EZBake::Config[:main_namespace] %> \
          --config ${CONFIG} --bootstrap-config ${BOOTSTRAP_CONFIG} \
          --restart-file "${restartfile}" \
+         ${TK_ARGS} \
```
sarameisburger: @Sharpie added!

mwaggett:
🙈 this looks reasonable to me, I will will give you some unofficial, not-especially-confident-because-I'm-bad-at-vanagon thumbs.

resources/puppetlabs/lein-ezbake/template/global/ext/redhat/init.suse.erb
```diff
@@ -43,6 +43,7 @@ config=$CONFIG
 PATH=/sbin:/usr/sbin:/bin:/usr/bin
 JARFILE="<%= EZBake::Config[:uberjar_name] %>"
 JAVA_ARGS="${JAVA_ARGS} -cp ${INSTALL_DIR}/${JARFILE} clojure.main -m <%= EZBake::Config[:main_namespace] %> --config ${CONFIG} -b ${BOOTSTRAP_CONFIG}"
+TK_ARGS="${TK_ARGS}"
```
underscorgan: I don't think you need this, I believe this will be defined by the defaults file

resources/puppetlabs/lein-ezbake/template/foss/ext/default.erb
```diff
@@ -8,6 +8,9 @@ JAVA_BIN="/usr/bin/java"
 # Modify this if you'd like to change the memory allocation, enable JMX, etc
 JAVA_ARGS="<%= EZBake::Config[:java_args] %>"
 
+# Modify this if you'd like TrapperKeeper specific arguments
+TK_ARGS="<%= EZBake::Config[:tk_args] %>"
```
underscorgan: You'll also want these changes in template/pe/ext/default.erb

resources/puppetlabs/lein-ezbake/template/pe/ext/default.erb
```diff
@@ -8,6 +8,9 @@ JAVA_BIN="/opt/puppetlabs/server/bin/java"
 # Modify this if you'd like to change the memory allocation, enable JMX, etc
 JAVA_ARGS="<%= EZBake::Config[:java_args] %>"
 
+# Modify this if you'd like TrapperKeeper specific arguments
+TK_ARGS="<%= EZBake::Config[:tk_args] %>"
```
sarameisburger: @underscorgan added here and removed from the other file!

PR #489, opened 2018-05-17
(EZ-124) Use consistent timestamps

underscorgan:
We use timestamps for two separate parts of the `stage` action. Since
the timestamps only record the time to the minute, this only causes
issues when the timing is just right, but when it does it leads to
a mismatch in package version and the directory the packages are shipped
to.

This commit updates the stage action to generate the timestamp once and
pass it through to the functions that ultimately end up setting the
release field in ezbake.rb and tagging the git repo so we can be sure
we're using the same timestamp for both.

### DIFF
--- a/src/puppetlabs/ezbake/core.clj
+++ b/src/puppetlabs/ezbake/core.clj
@@ -446,13 +446,13 @@ Additional uberjar dependencies:
       (str/replace "-" ".")))
 
 (defn generate-git-tag-from-version
-  [lein-version]
+  [lein-version timestamp]
   {:pre [(string? lein-version)]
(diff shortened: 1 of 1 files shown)


underscorgan:
@mwaggett this is all in the stage command, which makes it easier to test while doing development.

I did kick off a build for good measure, and it worked https://jenkins-master-prod-1.delivery.puppetlabs.net/job/platform_various-packaging-jobs_packaging-os-clj_lein-ezbake-generic/795

I was also able to test over the minute boundary and make sure it was doing what I expected it to be doing, see https://github.com/underscorgan/ezbake/tree/scratchspace

PR #498, opened 2018-08-21
(maint) packaging-as-a-gem improvements

mwaggett:
cherry-picked the merge commit from master

### DIFF
--- a/src/puppetlabs/ezbake/core.clj
+++ b/src/puppetlabs/ezbake/core.clj
@@ -847,7 +847,7 @@ Additional uberjar dependencies:
 (defmethod action "build"
   [_ lein-project build-target]
   (action "stage" lein-project build-target)
-  (exec/exec "bundle" "install" :dir staging-dir)
+  (exec/exec "bundle" "install" "--path" ".bundle/gems" "--binstubs" ".bundle/bin" :dir staging-dir)
   (let [downstream-job nil
         rake-call ["bundle" "exec" "rake" "pl:jenkins:trigger_build_local_auth"]]
     (exec/lazy-sh rake-call {:dir staging-dir})))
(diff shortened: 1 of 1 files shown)


src/puppetlabs/ezbake/core.clj
```diff
@@ -847,22 +847,25 @@ Additional uberjar dependencies:
 (defmethod action "build"
   [_ lein-project build-target]
   (action "stage" lein-project build-target)
+  (exec/exec "bundle" "install" :dir staging-dir)
```
demophoon: Do we also need to specify a `--path` flag here as to install to a self contained gem path?

mwaggett: Is that not what `:dir staging-dir` is for?

mwaggett: it might not be

mwaggett: I am bad at Clojure

underscorgan: @mwaggett `:dir staging-dir` doesn't have anything to do with the `--path` like you have in other bundle installs, the `:dir` specifies where the command runs and `--path` is the path bundler will install things into. Probably missed that in the other PR.

src/puppetlabs/ezbake/core.clj
```diff
@@ -847,22 +847,25 @@ Additional uberjar dependencies:
 (defmethod action "build"
   [_ lein-project build-target]
   (action "stage" lein-project build-target)
+  (exec/exec "bundle" "install" "--path" ".bundle/gems" "--binstubs" ".bundle/bin" :dir staging-dir)
```
mwaggett: @underscorgan @demophoon updated (will cherry-pick up to master after this is merged).

underscorgan: have you run a test build since this change?

PR #502, opened 2018-08-31
(RE-109140) Improve errors in ezbake builds

underscorgan:


underscorgan:
With a bad auth token:
```

Error triggering job: https://jenkins-master-prod-1.delivery.puppetlabs.net/job/platform_various-packaging-jobs_packaging-os-clj_lein-ezbake-generic
HTTP response: HTTP/1.1 403 Forbidden

You need to pass the environment variable JENKINS_USER_AUTH
It should be in the format <LDAP username>:<access token>
To find your access token, go to http://<jenkins-url>/me/configure
These jobs also are configured with an authentication token
that you can use instead of your personal token. To find the
job authentication token see the 'Build Triggers' section of
https://jenkins-master-prod-1.delivery.puppetlabs.net/job/platform_various-packaging-jobs_packaging-os-clj_lein-ezbake-generic/configure. In this case, JENKINS_USER_AUTH should
be set to only the Authentication Token.
```

Missing PE_VER:
```
Error triggering job: https://cinext-jenkinsmaster-enterprise-prod-1.delivery.puppetlabs.net/job/enterprise_various-packaging-jobs_packaging-os-clj_lein-ezbake-generic
Requires PE_VER be set!
```

PR #503, opened 2018-08-31
(RE-109140) improve error reporting

underscorgan:


underscorgan:
cherry-picks from master

PR #512, opened 2018-12-03
(EZ-128) Add support for specifying numeric uid/gid in rpm packages

underscorgan:
For RPM packaging, we have a reserved uid/gid for the 'puppet' user. To
enable us to use the reserved uid/gid for the puppet user, ezbake needs
to support an option for that. The new option is `numeric-uid-gid`,
which can be set in the `:lein-ezbake` section of your project.clj.

This option is a noop for deb packaging. According to the debian policy
guide
(https://www.debian.org/doc/debian-policy/ch-opersys.html#uid-and-gid-classes)
reserved uids and gids need to be set in the `base-passwd` package.

### DIFF
--- a/resources/puppetlabs/lein-ezbake/template/global/ext/redhat/preinst.erb
+++ b/resources/puppetlabs/lein-ezbake/template/global/ext/redhat/preinst.erb
@@ -2,12 +2,25 @@
 # install.sh source based installation methodology.
 #
 # Add <%= EZBake::Config[:group] %> group
-getent group <%= EZBake::Config[:group] %> > /dev/null || \
+if getent group <%= EZBake::Config[:group] %> > /dev/null; then
+  : noop
+<% unless EZBake::Config[:numeric_uid_gid].nil? -%>
+elif ! getent group <%= EZBake::Config[:numeric_uid_gid] %> > /dev/null; then
+  groupadd -r --gid <%= EZBake::Config[:numeric_uid_gid] %> <%= EZBake::Config[:group] %> || :
+<% end -%>
+else
   groupadd -r <%= EZBake::Config[:group] %> || :
+fi
+
 # Add <%= EZBake::Config[:user] %> user
 if getent passwd <%= EZBake::Config[:user] %> > /dev/null; then
   usermod --gid <%= EZBake::Config[:group] %> --home %{_app_data} \
   --comment "<%= EZBake::Config[:project] %> daemon" <%= EZBake::Config[:user] %> || :
+<% unless EZBake::Config[:numeric_uid_gid].nil? -%>
+  elif ! getent passwd <%= EZBake::Config[:numeric_uid_gid] %> > /dev/null; then
+  useradd -r --uid <%= EZBake::Config[:numeric_uid_gid] %> --gid <%= EZBake::Config[:group] %> --home %{_app_data} --shell $(which nologin) \
+    --comment "<%= EZBake::Config[:project] %> daemon"  <%= EZBake::Config[:user] %> || :
+<% end %>
 else
   useradd -r --gid <%= EZBake::Config[:group] %> --home %{_app_data} --shell $(which nologin) \
     --comment "<%= EZBake::Config[:project] %> daemon"  <%= EZBake::Config[:user] %> || :
--- a/resources/puppetlabs/lein-ezbake/template/global/ext/redhat/preinst.erb
+++ b/resources/puppetlabs/lein-ezbake/template/global/ext/redhat/preinst.erb
@@ -2,25 +2,21 @@
 # install.sh source based installation methodology.
 #
 # Add <%= EZBake::Config[:group] %> group
-if getent group <%= EZBake::Config[:group] %> > /dev/null; then
-  : noop
-<% unless EZBake::Config[:numeric_uid_gid].nil? -%>
-elif ! getent group <%= EZBake::Config[:numeric_uid_gid] %> > /dev/null; then
-  groupadd -r --gid <%= EZBake::Config[:numeric_uid_gid] %> <%= EZBake::Config[:group] %> || :
+<% if EZBake::Config[:numeric_uid_gid].nil? -%>
+getent group <%= EZBake::Config[:group] %> >/dev/null || groupadd --system --force <%= EZBake::Config[:group] %>
+<% else -%>
+getent group <%= EZBake::Config[:group] %> >/dev/null || groupadd --system --force --gid <%= EZBake::Config[:numeric_uid_gid] %> <%= EZBake::Config[:group] %>
 <% end -%>
-else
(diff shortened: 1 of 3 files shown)


underscorgan:
Using a build of ezbake off of this branch I built puppetserver with no configuration changes and with `:numeric-uid-gid` specified. These both worked as expected.

resources/puppetlabs/lein-ezbake/template/global/ext/redhat/preinst.erb
```diff
@@ -2,12 +2,25 @@
 # install.sh source based installation methodology.
 #
 # Add <%= EZBake::Config[:group] %> group
-getent group <%= EZBake::Config[:group] %> > /dev/null || \
+if getent group <%= EZBake::Config[:group] %> > /dev/null; then
+  : noop
+<% unless EZBake::Config[:numeric_uid_gid].nil? -%>
+elif ! getent group <%= EZBake::Config[:numeric_uid_gid] %> > /dev/null; then
```
mwaggett: Is this checking if the specified gid already exists? If so, and the gid *does* exist, should we fail rather than creating the group anyway (at a different gid??)?

(same for uid below)

resources/puppetlabs/lein-ezbake/template/global/ext/redhat/preinst.erb
```diff
@@ -2,12 +2,25 @@
...
   groupadd -r <%= EZBake::Config[:group] %> || :
+fi
+
 # Add <%= EZBake::Config[:user] %> user
 if getent passwd <%= EZBake::Config[:user] %> > /dev/null; then
   usermod --gid <%= EZBake::Config[:group] %> --home %{_app_data} \
   --comment "<%= EZBake::Config[:project] %> daemon" <%= EZBake::Config[:user] %> || :
+<% unless EZBake::Config[:numeric_uid_gid].nil? -%>
+  elif ! getent passwd <%= EZBake::Config[:numeric_uid_gid] %> > /dev/null; then
+  useradd -r --uid <%= EZBake::Config[:numeric_uid_gid] %> --gid <%= EZBake::Config[:group] %> --home %{_app_data} --shell $(which nologin) \
```
mwaggett: Maybe want to DRY this up with the command in the `else` case below?

resources/puppetlabs/lein-ezbake/template/global/ext/redhat/preinst.erb
underscorgan: yes, that is checking if the specified gid already exists. You never want rpm scriptlets to exit with a non-zero exit code, so we don't want this to fail, and it would be much better to end up with the user or group added without the specified uid/gid than end up with the user or group not being added at all.

From the [Fedora Packaging Guidelines](https://fedoraproject.org/wiki/Packaging:Scriptlets?rd=Packaging:ScriptletSnippets):
> All scriptlets MUST exit with the zero exit status. Because RPM in its default configuration does not execute shell scriptlets with the -e argument to the shell, excluding explicit exit calls (frowned upon with a non-zero argument!), the exit status of the last command in a scriptlet determines its exit status. Most commands in the snippets in this document have a "|| :" appended to them, which is a generic trick to force the zero exit status for those commands whether they worked or not. Usually the most important bit is to apply this to the last command executed in a scriptlet, or to add a separate command such as plain ":" or "exit 0" as the last one in a scriptlet. Note that depending on the case, other error checking/prevention measures may be more appropriate.

>Non-zero exit codes from scriptlets can break installs/upgrades/erases such that no further actions will be taken for that package in a transaction (see #Scriptlet Ordering below), which may for example prevent an old version of a package from being erased on upgrades, leaving behind duplicate rpmdb entries and possibly stale, unowned files on the filesystem. There are some cases where letting the transaction to proceed when some things in scriptlets failed may result in partially broken setup. It is however often limited to that package only whereas letting a transaction to proceed with some packages dropped out on the fly is more likely to result in broader system wide problems.

mwaggett: ahh okay, that makes sense! Should we at least warn in that case then? Seems like someone might want to know if the behavior they expect is not what's actually happening?

underscorgan: No, I don't think we should warn them. Not warning them seems inline with the suggested soft-static uid/gid implementation from Fedora (https://fedoraproject.org/wiki/Packaging:UsersAndGroups#Soft_static_allocation).

I don't have a good resource for why warning feels wrong, but it feels like we should be silent in this case. I think that's what most other packages do (since it's soft static, not hard static), and in general I don't assume that anyone will be watching the installation happen to see the logs. Mostly, nothing in the package is going to break or require any action in the case it got set to a dynamically allocated uid, but something may end up scraping logs and alerting on a warning.

resources/puppetlabs/lein-ezbake/template/global/ext/redhat/preinst.erb
```diff
@@ -2,15 +2,24 @@
...
 else
-  useradd -r --gid <%= EZBake::Config[:group] %> --home %{_app_data} --shell $(which nologin) \
-    --comment "<%= EZBake::Config[:project] %> daemon"  <%= EZBake::Config[:user] %> || :
+  useradd_options=('--system' '--gid' '<%= EZBake::Config[:group] %>' '--home' '%{_app_data}' '--shell' "$(which nologin)" '--comment' '<%= EZBake::Config[:project] %> daemon')
+<% unless EZBake::Config[:numeric_uid_gid].nil? -%>
+  if ! getent passwd <%= EZBake::Config[:numeric_uid_gid] %> > /dev/null; then
+    useradd_options+=('--uid' '<%= EZBake::Config[:numeric_uid_gid] %>')
+  fi
+<% end -%>
+  useradd "${useradd_options[@]}" <%= EZBake::Config[:user] %> || :
```
underscorgan: I neither love nor hate this change, what do y'alls think 🤷‍♀️

shrug: I think I like this OK, honestly?

mwaggett: ^same.

Can you explain why you have to check if the uid/gid exists already when adding a user, but not when adding a group?

underscorgan: We're running groupadd with `--force`, so if we try to create it with a gid that already exists it will pick a different one ([manpage](https://linux.die.net/man/8/groupadd)). `useradd` doesn't support that functionality.

mwaggett: cool, thanks!

PR #516, opened 2018-12-21
(SERVER-2399) List JAVA_ARGS_CLI in default file

justinstoller:
Puppet Server at least provides cli tools in addition to ezbake's
provided `start`, `stop`, `reload`. Those tools also need a way to
specify JAVA_ARGS but have Very different needs than the service. Puppet
Server therefore uses JAVA_ARGS_CLI to differentiate between the one
off CLI tools and the service.

This patch adds that JAVA_ARGS_CLI to the default file that ezbake
provides.

### DIFF
--- a/resources/puppetlabs/lein-ezbake/staging-templates/ezbake.rb.mustache
+++ b/resources/puppetlabs/lein-ezbake/staging-templates/ezbake.rb.mustache
@@ -71,6 +71,7 @@ module EZBake
                           },
       :main_namespace => '{{{main-namespace}}}',
       :java_args      => '{{{java-args}}}',
+      :java_args_cli  => '{{{java-args-cli}}}',
       :tk_args        => '{{{tk-args}}}',
       :start_after    => [{{{start-after}}}],
       :replaces_pkgs  => {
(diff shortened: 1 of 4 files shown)


justinstoller:
I don't know if this is the correct solution or if this is the correct branch to target (this goes back at least until the 5 series).

sarameisburger:
When will JAVA_ARGS_CLI be called? My ezbake knowledge is rusty but a while back I added TK_ARGS https://github.com/puppetlabs/ezbake/pull/479 if that helps as an example

Magisus:
Example of use in puppetserver's custom CLI tools: https://github.com/puppetlabs/puppetserver/blob/074fe9701ceff3e62cb17889ec768840214d143b/resources/ext/cli/irb.erb

Magisus:
Looking at yours, I think this might be missing the bit in `ezbake/core.clj` where it actually gets added to the config map.

mwaggett:
yeah, based on Sara's, this seems to be missing some things.

When you think you're ready, can you build with this branch and confirm that it does what you want?

justinstoller:
Okay, so until there is an update to the core.clj the `EZBake::Config[:java_args_cli]` line bit will always return `nil` and can't have values set in a project.clj, if I understand correctly? I'll update this accordingly and test it in a package.

NB: We don't actually have any desire to configure the defaults right now (though maybe we want to reconsider that) the primary point of this was to have the variable in the defaults file so it was more discoverable if users wanted to change it.

justinstoller:
Updated core.clj, building a package now.

justinstoller:
Okay, I manually tested that this works as expected with http://builds.delivery.puppetlabs.net/puppetserver/6.1.1.SNAPSHOT.2019.01.08T1250/artifacts/el/7/puppet6/x86_64/ for anyone that wants to double check.

PR #518, opened 2019-01-08
(PE-24606) use luseradd instead of useradd for ldap reboot issue

sarameisburger:
Some customers have a requirement to have all locally created users exist in LDAP. Because the same users would be created on every system, the local users for Puppet on a new master already exist in LDAP. The check is performed, the users are determined to already exist, and then during a reboot the systemd-tmpfiles-setup fails due to ldap dependencies of the puppet userid/groups

### DIFF
--- a/resources/puppetlabs/lein-ezbake/template/global/Rakefile
+++ b/resources/puppetlabs/lein-ezbake/template/global/Rakefile
@@ -1,6 +1,5 @@
 require 'rake'
 require './ezbake.rb'
-require 'libuser'
 
 puts "EZBAKE PROJECT NAME: #{EZBake::Config[:project]}"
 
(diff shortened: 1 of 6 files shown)


shrug:
Probably build and then test installs on all of our supported masters - so the most recent several ubuntu LTS, centos 6 and 7, debians, sles 11 and 12, recent debians....

shrug:
We may need to think about the other commands - usermod, groupadd, etc as well. We should work up a test environment.

shrug:
Also I think we'll need to add a dependency on the libuser package which provides the commands, and update all of them - luseradd, lusermod, lgroupadd, etc...

sarameisburger:
Great. Fun. Love it.

sarameisburger:
Ahh ok, ya wasn't sure so I just threw it in there

shrug:
lgtm now, does it blend?

sarameisburger:
Tested with el-7 and bionic 👍

mwaggett:
perhaps worth testing on a sles as well?
but otherwise, lgtm!

PR #524, opened 2019-01-14
(maint) We only support JRE 8 in some of our products

justinstoller:


justinstoller:
No rush on this, just a mistake I noticed while going through some support output.

PR #529, opened 2019-03-01
(RE-12082) Revert libuser changes

mwaggett:
This commit reverts the changes made to address PE-24606. We encountered
installation failure when bumping puppetserver to a version of ezbake with
these changes. There were several issues with this work, the largest of which
being the fact that there is no libuser package for sles and there is a bug in
libuser1, a dependency of libuser, for debian 8 (jessie) and ubuntu 16.04
(xenial) (https://bugs.launchpad.net/ubuntu/+source/system-config-samba/+bug/1718202).
Working around these issues (i.e. only using libuser commands on rhel systems)
is possible, but there are several unused code paths in this repo that are
making this hard to debug. The plan is to remove those paths before retrying
this work.

Revert "add libuser dependency in the correct spot"
This reverts commit c4e988049b128e8bd4f44f59e0cb8969bef3f028.

Revert "add libuser and update commands"
This reverts commit 16c5334a51cb9f7021e74e712814dd330668be63.

Revert "(PE-24606) use luseradd instead of useradd for ldap reboot issue"
This reverts commit 69727d1855f86e00f154ca6e1fab895d7e94be87.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 1.9.7 - 2019-03-04
+This is a bugfix release.
+
+Bugfix:
+  * (RE-12082) Return to using `adduser` commands instead of `libuser` commands.
+    This reverts the work done for PE-24606 in the 1.9.5 release.
+
 ## 1.9.6 - 2019-02-25
 This is a maintenance release.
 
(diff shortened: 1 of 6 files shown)


mwaggett:
Built all packages: https://jenkins-master-prod-1.delivery.puppetlabs.net/job/platform_various-packaging-jobs_packaging-os-clj_lein-ezbake-generic/1541/
Confirmed that I can install and start puppetserver on el 7, sles 12, and ubuntu 16.04.

PR #530, opened 2019-03-07
(RE-12093) Remove unused files

mwaggett:
This commit removes the rpm spec and debian control templates that were
previously used to build packages. We now use fpm to build, so these are no
longer needed.


mwaggett:
Built, installed, and started puppetserver on el 7 and ubuntu 16.04.

resources/puppetlabs/lein-ezbake/staging-templates/project_data.yaml.mustache
```diff
@@ -45,7 +43,6 @@ templates:
   - ext/redhat/init.erb
   - ext/redhat/init.suse.erb
   - ext/debian/changelog.erb
-  - ext/debian/control.erb
   - ext/debian/dirs.erb
```
underscorgan: mm, we might be able to get rid of a few more of these deb files, lemme check

mwaggett: yeah, I think there's a lot more to get rid of, but this is what I got to testing last week, so I wanted to get it in before pivoting to higher-priority stuff.
If there are files in that list that are already gone or things you're pretty confident can go, I'll add that to this PR, but otherwise, I think this work is less important than some of the other things we have in flight.

underscorgan: Yeah, there are a few very related files (in the same directory, the other debian build files that fpm doesn't use). It seems like if we don't clean those now we will never do it, i'm running a build now just to test.

mwaggett: okay cool, yeah, related files seem like a good thing to include. thanks for testing that!

underscorgan:
@mwaggett I pushed up a commit with the additional files that could be removed, tested installing and starting server on el7/xenial with those changes.

PR #531, opened 2019-03-12
(maint) Update changelog for 2.0.0 release

sarameisburger:


CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 1.9.8 - 2019-03-12
```
mwaggett: This should be a 2.0.0 release, since removing those files is backwards-incompatible, according to @underscorgan.

sarameisburger: ok

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 2.0.0 - 2019-03-12
+This is a major feature release.
```
sarameisburger: @mwaggett updated. Wasn't sure if this is what we want to call it tho

mwaggett: hmm, yeah. maybe `backwards-incompatible release`? since it's not really a new feature.. ¯\_(ツ)_/¯

underscorgan:
hmm yeah before we do the 2.0 release we should at least remove the task that is definitely not still functional since we removed those packaging artifacts. We might want to see if there's anything else obvious that needs to happen.

underscorgan:
https://github.com/puppetlabs/ezbake/pull/532 at least removes legacy-build

CHANGELOG.md
sarameisburger: @underscorgan is there a specific wording i should be using? I looked at https://semver.org/ but wasn't sure

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
...
+
+Feature:
+  * (RE-12093) Remove old rpm and deb artifacts previously used to build packages.
+  * Remove legacy build action that used the removed artifacts. Packages can only
+  be built using FPM.
+
```
sarameisburger: @underscorgan updated

CHANGELOG.md
underscorgan: mm, yeah, probably backwards-incompatible release.

CHANGELOG.md
underscorgan: can you switch legacy build to `legacy-build` for extra clarity?

sarameisburger:
@underscorgan @mwaggett updated!

PR #536, opened 2019-05-24
(RE-12428) Add support for redhatfips-7

sarameisburger:
Needs testing

### DIFF
--- a/resources/puppetlabs/lein-ezbake/template/pe/ext/build_defaults.yaml
+++ b/resources/puppetlabs/lein-ezbake/template/pe/ext/build_defaults.yaml
@@ -6,7 +6,7 @@ packager: 'puppet'
 gpg_key: '7F438280EF8D349F'
 sign_tar: FALSE
 # a space separated list of mock configs
-final_mocks: 'pupent-el6-x86_64 pupent-el7-x86_64 pupent-el8-x86_64 pupent-sles12-x86_64'
+final_mocks: 'pupent-el6-x86_64 pupent-el7-x86_64 pupent-el8-x86_64 pupent-redhatfips-7-x86_64 pupent-sles12-x86_64'
 yum_host: 'yum.puppetlabs.com'
 yum_repo_path: '/opt/repository/yum/'
 build_gem: FALSE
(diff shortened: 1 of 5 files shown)


sarameisburger:
Successful build of puppetserver https://jenkins-master-prod-1.delivery.puppetlabs.net/job/platform_various-packaging-jobs_packaging-os-clj_lein-ezbake-generic/1803/

resources/puppetlabs/lein-ezbake/template/pe/ext/build_defaults.yaml
```diff
@@ -6,7 +6,7 @@ packager: 'puppet'
 gpg_key: '7F438280EF8D349F'
 sign_tar: FALSE
 # a space separated list of mock configs
-final_mocks: 'pupent-el6-x86_64 pupent-el7-x86_64 pupent-el8-x86_64 pupent-sles12-x86_64'
+final_mocks: 'pupent-el6-x86_64 pupent-el7-x86_64 pupent-el8-x86_64 pupent-redhatfips-7-x86_64 pupent-sles12-x86_64'
```
mwaggett: I know packaging (sort of) uses these values, but I'm not totally sure how/where. The `pupent` thing is something I've only ever seen in build_defaults, so I'm not totally sure if this is formatted correctly or not (e.g. should this be `pupent-redhatfips7-x86_64` (no space between platform and version) to match the rest??), or whether there needs to be some additional setup for this.

sarameisburger: Yeah I don't really know why they're formatted differently from the foss build_defaults but will update!

shrug: Legacy from old build infra that we no longer use, but packaging still parses and uses the names. It's from the old mock builds: https://github.com/rpm-software-management/mock/wiki

sarameisburger: Ah gotcha. Updated.

PR #541, opened 2019-07-10
 (PDB-4373) ezbake.rb: insert complete, escaped ruby literals

rbrw:


rbrw:
I went ahead and included "quite a bit" in this pr, but I'm more than happy to dial it back if that's preferable for now.  And with respect to the travis work, I just did the fastest thing and copied bits of the infrastructure we use in pdb (for example, we manage the jdk ourselves because travis changes available versions depending on the ubuntu version, etc. and we didn't want to have two different ways to manage versions).

rbrw:
Oh, and fwiw, one other reason pdb handles things that way is that all else equal, it's nice to have tools that are fairly independent of the (CI) system involved.  i.e. these commands should work on any relevant system (here it's debianish linux and osx), they don't require root, and we can also use them (and debug them) locally.

And at least regarding the jdks in travis, iirc some of the complexity was unavoidable for pdb since it needed to test openjdk8 and openjdk11 on linux and osx, and travis didn't support that directly.

In any case, I'm completely fine doing something simpler for ezbake if that's preferable.  This was just "easy" given what we already had.

underscorgan:
Would love to get review from someone with more clojure knowledge but afaict this seems reasonable? I looked at generated `ezbake.rb` for puppetserver before and after this change and the only differences were changes from `"` to `'` which should be fine.

rbrw:
Oh, and I left the "notifications" commented out because I didn't know if we wanted them going anywhere.  If we do, I can update it, otherwise I can take them out in the final version.

src/puppetlabs/ezbake/core.clj
```diff
@@ -189,9 +189,38 @@
...
+  Long (as-ruby-literal [s] (str s))
+  BigInteger (as-ruby-literal [s] (str s))
+  clojure.lang.BigInt (as-ruby-literal [s] (str s))
+
+  String
+  (as-ruby-literal [s]
+    (let [esc (-> s
+                  (str/replace "\\" "\\\\")
+                  (str/replace "'" "\\'"))]
+      (str "'" esc "'")))
```
justinstoller: Some of the ruby values in `ezbake.rb.moustache` have single quotes and some have double quotes. Double quotes have different interpolation rules than single quotes. Do you know if that matters?

...iow, do we ever pass set a string in our clojure ezbake configuration that we expect to be interpolated within Ruby?

justinstoller: Ahh, that behavior is intentional (finally reading the commit).

justinstoller: I think the concern is valid (though I hope we never do that anywhere) and slightly different than the point of the PR. I've double checked and I don't think[pe-]puppetserver isn't using such tomfoolery.

shrug:
It looks like this needs a rebase, happy to merge after it's sorted

rbrw:
OK, rebased.  Thanks.

PR #542, opened 2019-07-15
(maint) Add docker container for ezbake builds

underscorgan:
This commit adds artifacts needed to build and test an ezbake container.
This should help with development and testing, both internal to Puppet
and for external contributors.

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -26,13 +26,13 @@ else
 endif
 
 build: prep
-	  @docker build \
-			--pull \
-			--build-arg vcs_ref=$(vcs_ref) \
-			--build-arg build_date=$(build_date) \
-			--build-arg version=$(version) \
-			--file ezbake/$(dockerfile) \
-			--tag $(NAMESPACE)/ezbake:$(version) $(pwd)/..
+	@docker build \
+		--pull \
+		--build-arg vcs_ref=$(vcs_ref) \
+		--build-arg build_date=$(build_date) \
+		--build-arg version=$(version) \
+		--file ezbake/$(dockerfile) \
+		--tag $(NAMESPACE)/ezbake:$(version) $(pwd)/..
 ifeq ($(IS_LATEST),true)
 	@docker tag $(NAMESPACE)/ezbake:$(version) $(NAMESPACE)/ezbake:latest
 endif
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -1,0 +1,51 @@
+NAMESPACE ?= puppet
+git_describe := $(shell git describe)
+vcs_ref := $(shell git rev-parse HEAD)
+build_date := $(shell date -u +%FT%T)
+hadolint_available := $(shell hadolint --help > /dev/null 2>&1; echo $$?)
+hadolint_command := hadolint --ignore DL3008 --ignore DL3018 --ignore DL3028 --ignore DL4000 --ignore DL4001
+hadolint_container := hadolint/hadolint:latest
+pwd := $(shell pwd)
+export BUNDLE_PATH = $(pwd)/.bundle/gems
+export BUNDLE_BIN = $(pwd)/.bundle/bin
+export GEMFILE = $(pwd)/Gemfile
+
+version := $(shell echo $(git_describe) | sed 's/-.*//')
+dockerfile := Dockerfile
+
+prep:
+	@git fetch --unshallow 2> /dev/null ||:
+	@git fetch origin 'refs/tags/*:refs/tags/*'
+
+lint:
+ifeq ($(hadolint_available),0)
+	@$(hadolint_command) ezbake/$(dockerfile)
+else
+	@docker pull $(hadolint_container)
+	@docker run --rm -v $(PWD)/ezbake/$(dockerfile):/Dockerfile -i $(hadolint_container) $(hadolint_command) Dockerfile
+endif
+
+build: prep
+	  @docker build \
+			--pull \
+			--build-arg vcs_ref=$(vcs_ref) \
+			--build-arg build_date=$(build_date) \
+			--build-arg version=$(version) \
+			--file ezbake/$(dockerfile) \
+			--tag $(NAMESPACE)/ezbake:$(version) $(pwd)/..
+ifeq ($(IS_LATEST),true)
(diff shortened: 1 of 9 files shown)


docker/Makefile
```diff
@@ -0,0 +1,51 @@
...
+lint:
+ifeq ($(hadolint_available),0)
+	@$(hadolint_command) ezbake/$(dockerfile)
+else
+	@docker pull $(hadolint_container)
+	@docker run --rm -v $(PWD)/ezbake/$(dockerfile):/Dockerfile -i $(hadolint_container) $(hadolint_command) Dockerfile
+endif
+
+build: prep
+	  @docker build \
```
mwaggett: (nitpick) This looks slightly misaligned with everything else

docker/Makefile
```diff
@@ -0,0 +1,51 @@
...
+			--pull \
+			--build-arg vcs_ref=$(vcs_ref) \
+			--build-arg build_date=$(build_date) \
+			--build-arg version=$(version) \
+			--file ezbake/$(dockerfile) \
+			--tag $(NAMESPACE)/ezbake:$(version) $(pwd)/..
+ifeq ($(IS_LATEST),true)
+	@docker tag $(NAMESPACE)/ezbake:$(version) $(NAMESPACE)/ezbake:latest
+endif
```
mwaggett: What's the benefit of tagging as part of `build` as opposed to `publish`?

docker/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+    ```
+    docker run --rm --volume $(PWD)/output:/output --volume /workspace/test --volume /workspace/test-resources --volume /workspace/target --volume /workspace/tmp puppet/ezbake
+    ```
+    
+    b. Only mount in needed files/directories
+
+    This is faster than mounting in everything and using anonymous volumes, but it'll take more effort
+    to get the correct things mounted in. This will vary some between projects, but you can get a pretty
+    good idea of what is needed by looking at the `.gitignore` or `.dockerignore` files.
```
mwaggett: I assume the things in those files are things you DON'T need? That could be made clearer..

docker/distelli-manifest.yml
```diff
@@ -0,0 +1,9 @@
+release-engineering/ezbake:
```
mwaggett: Is this something we want, now that Pipelines/Distelli is EOL?

docker/ezbake/docker-entrypoint.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+
+set -e
+
+if [ -n "$EZBAKE_REPO" ]; then
+  git clone $EZBAKE_REPO /ezbake
+  cd /ezbake
+  if [ -n "$EZBAKE_BRANCH" ]; then
+    git checkout origin/$EZBAKE_BRANCH
```
mwaggett: Why does `origin` need to be specified? I'm wondering if the BRANCH/REF vars can be consolidated.

docker/ezbake/docker-entrypoint.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+
+set -e
+
+if [ -n "$EZBAKE_REPO" ]; then
```
mwaggett: What happens if `$EZBAKE_REPO` is unset? It looks like things will continue, but I'm not sure how that's supposed to work..

docker/README.md
mwaggett: oh wait, I see - the .dockerignore says to ignore everything _except_ those things.. Maybe only reference that, instead of the .gitignore? since they're following different patterns here.

docker/Makefile
underscorgan: The image is already tagged as part of build, this just adds an additional tag at build time. This is a standard pattern we have for pupperware containers, and tagging at build time is very useful for local investigations/tests

docker/README.md
underscorgan: I included both of them in case the `.dockerignore` file doesn't exist, since you may be building a project (or a branch) that doesn't have a corresponding container.

docker/distelli-manifest.yml
underscorgan: Yes. We haven't figured out our migration path yet for pupperware, and will migrate everything once we figure that out. I don't want that to block other containerization work

docker/ezbake/docker-entrypoint.sh
underscorgan: the build will pull whatever version of ezbake specified in project.clj from clojars/artifactory

underscorgan: and yes, that's how it's supposed to work

docker/ezbake/docker-entrypoint.sh
underscorgan: I was having issues getting branches to actually be checked out without specifying origin, but I'll test again.

docker/Makefile
underscorgan: updated

docker/README.md
mwaggett: ookay!

docker/distelli-manifest.yml
mwaggett: that's fair, just wanted to check 😊

docker/ezbake/docker-entrypoint.sh
underscorgan: yep, can confirm that when i pass a branch name to REF it doesn't actually check out a different branch/commit, just whatever the default is

docker/ezbake/docker-entrypoint.sh
mwaggett: ahhh okay, I see.

And to make sure I'm understanding correctly: if `$PROJECT_REPO` is unset, it just assumes that your project is already mounted into the `/workspace` directory and tries to continue with that?

docker/ezbake/docker-entrypoint.sh
mwaggett: weird!

docker/ezbake/docker-entrypoint.sh
underscorgan: yep!

PR #548, opened 2019-08-15
(SERVER-2602) Do not manage egd

justinstoller:
This setting is of little value in Java 8+ and we've seen no difference
anecdotally between explicitly configuring /dev/urandom as a source vs
using the JVMs defaults.

We will be explicitly setting this to /dev/random for FIPS enabled
builds, however we will do so via the JAVA_ARGS variable. FOSS users who
would like to manage their egd source should use JAVA_ARGS as well.

### DIFF
--- a/resources/puppetlabs/lein-ezbake/template/global/ext/cli/foreground.erb
+++ b/resources/puppetlabs/lein-ezbake/template/global/ext/cli/foreground.erb
@@ -28,7 +28,6 @@ if [ -e "$cli_defaults" ]; then
 fi
 
 COMMAND="${JAVA_BIN} ${JAVA_ARGS} ${LOG_APPENDER} \
-         -Djava.security.egd=file:/dev/urandom \
          -cp "$CLASSPATH" \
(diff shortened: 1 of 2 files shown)


mwaggett:
I don't have much context on this, but seems reasonable to me!

Have you built puppetserver with this to confirm it works?

justinstoller:
> Have you built puppetserver with this to confirm it works?

No, I've updated in-place the start subcommand on Server 6.5.0 with this change. Which is where I got the "no difference [in performance] anecdotally". I'll do a quick build with this change.

justinstoller:
Built and confirmed with: http://builds.delivery.puppetlabs.net/puppetserver/6.5.1.SNAPSHOT.2019.08.15T1208/artifacts/el/8/puppet6/x86_64/

justinstoller:
btw, this is needed for lovejoy, but it's not time sensitive otherwise, don't let me get in the way of shipping the latest foss releases!

mwaggett:
@e-gris this can be merged whenever, since we'll need to do an ezbake release for it to be used anyway

PR #550, opened 2019-08-21
(maint) Prepare for 2.1.1 release

sarameisburger:


CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 2.1.1 - 2019-08-21
+This is a maintenance release.
+
+Bugfix:
+  * (SERVER-2602) Do not manage egd. For FIPS enabled builds explicitly set this to `/dev/random` via the `JAVA_ARGS` variable. FOSS users who would like to manage their egd source should use `JAVA_ARGS` as well.
```
sarameisburger: @justinstoller is this a backwards-incompatible change?

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
...
+  * Fix deprecation warning to suggest using `Pkg::Util::Git.fail_on_dirty_source`.
```
underscorgan: This is worded a little strangely. Maybe something more like 'Stop calling deprecated method `Pkg::Util::Version.fail_on_dirty_source`' ?

CHANGELOG.md
justinstoller: AFAICT, this change doesn't have an effect when running on Java >= 8.

CHANGELOG.md
sarameisburger: sounds good. i wasn't sure the best way to word it

sarameisburger: alrighty i updated it

CHANGELOG.md
sarameisburger: ahh gotcha. and is this description accurate? there were references to `dev/random` and `dev/urandom` in the PR

mwaggett:
```suggestion
  * (SERVER-2602) Do not manage egd. For FIPS-enabled builds, explicitly set this to `/dev/random` via the `JAVA_ARGS` variable. FOSS users who would like to manage their egd source should use `JAVA_ARGS` as well.
```

justinstoller: The middle sentence is very PE FIPS build specific and could probably be dropped. Something like: "Do not hard code egd in included subcommand files. Users that would like to manage egd source may do so via JAVA_ARGS (or their java.security file" I think contains the info that a FOSS user/consumer would want to know. (sorry, I should have written a better commit message)

sarameisburger: alrighty, updated!

PR #551, opened 2019-08-23
(maint) Extract name from bin files before converting to Ruby value

Magisus:
These bin files are also be JarFileEntries, and need to have their name
extracted before they can be converted into a Ruby string literal for
interpolation.


Magisus:
Does this make sense? This showed up when trying to build pe-puppetserver. Are there other contexts where we expect `bin-files` to NOT be JarFileEntries, or was this just an oversight?

Magisus:
https://github.com/puppetlabs/ezbake/blob/master/src/puppetlabs/ezbake/core.clj#L304-L306 makes it look like it's pretty likely these always come from jars.

PR #556, opened 2019-09-11
(maint) Use java11 for debian 10

underscorgan:


Magisus:
I built a new package with this and it appears to work great. The server also seems function on Debian 10, with Java 11, including reloading. If we're comfortable with this path forward, it works for me.

PR #566, opened 2020-01-29
Migrate pipelines to actions

underscorgan:
No changelog needed for this I don't think, since it's just a change in what tool we use to build the ezbake docker container.

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -52,8 +52,8 @@ push-readme:
 	@docker pull sheogorath/readme-to-dockerhub
 	@docker run --rm \
 		-v $(PWD)/README.md:/data/README.md \
-		-e DOCKERHUB_USERNAME="$(DISTELLI_DOCKER_USERNAME)" \
-		-e DOCKERHUB_PASSWORD="$(DISTELLI_DOCKER_PW)" \
+		-e DOCKERHUB_USERNAME="$(DOCKERHUB_USERNAME)" \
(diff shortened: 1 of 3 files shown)


docker/Makefile
```diff
@@ -1,16 +1,15 @@
...
-pwd := $(shell pwd)
-export BUNDLE_PATH = $(pwd)/.bundle/gems
-export BUNDLE_BIN = $(pwd)/.bundle/bin
-export GEMFILE = $(pwd)/Gemfile
+export BUNDLE_PATH = $(PWD)/.bundle/gems
+export BUNDLE_BIN = $(PWD)/.bundle/bin
+export GEMFILE = $(PWD)/Gemfile
 
-version := $(shell echo $(git_describe) | sed 's/-.*//')
+version ?= $(shell echo $(git_describe) | sed 's/-.*//')
```
e-gris: Why '?='  ?
Is overriding this useful?

underscorgan: When working on development for other containers I've used the ability to override `version` fairly regularly. So, not sure it's needed here, but it's at least a bit more consistent. `:=` is wrong anyways, and that was failing in CI (since you need to un-shallow the clone before you can `git describe`), so it would need to be `=`.

e-gris: Cool. Was mostly curious about the '?'; agree that the := would definitely be problematic

PR #568, opened 2020-02-07
(RE-12523) Don't set `TasksMax` in the service file for Debian 8

underscorgan:
This option isn't supported with the version of systemd that ships with
Debian 8, so comment out that line at install time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 This changelog adheres to [Keep a CHANGELOG](http://keepachangelog.com/).
 
 ## [Unreleased]
+Bugfix:
+  * (RE-12523) Don't set `TasksMax` in the Debian 8 service file since that option
+    is not supported by the included version of systemd.
 
 ## [2.1.5] - 2019-10-08
 This is a maintenance release.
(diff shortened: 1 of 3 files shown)


underscorgan:
Mmm, I think this is going to need some updates in the controller to have separate template extrapolation for jessie

underscorgan:
On jessie:
```
# https://tickets.puppetlabs.com/browse/EZ-129
# Prior to systemd v228, TasksMax was unset by default, and unlimited. Starting in 228 a default of '512'
# was implemented. This is low enough to cause problems for certain applications. In systemd 231, the
# default was changed to be 15% of the default kernel limit. This explicitly sets TasksMax to 4915,
# which should match the default in systemd 231 and later.
# See https://github.com/systemd/systemd/issues/3211#issuecomment-233676333
# Don't set TasksMax since the option doesn't exist
# TasksMax=4915
```

and on bionic:
```
# https://tickets.puppetlabs.com/browse/EZ-129
# Prior to systemd v228, TasksMax was unset by default, and unlimited. Starting in 228 a default of '512'
# was implemented. This is low enough to cause problems for certain applications. In systemd 231, the
# default was changed to be 15% of the default kernel limit. This explicitly sets TasksMax to 4915,
# which should match the default in systemd 231 and later.
# See https://github.com/systemd/systemd/issues/3211#issuecomment-233676333
TasksMax=4915
```

underscorgan:
@e-gris yeah, we do unfortunately. See https://github.com/puppetlabs/ezbake/blob/master/resources/puppetlabs/lein-ezbake/template/global/ext/debian/ezbake.service.erb#L31-L37

PR #572, opened 2020-05-11
(maint) Remove support for eol platforms

sarameisburger:
remove el-6 and ubuntu 16.04 xenial for foss and pe


sarameisburger:
@underscorgan does the foss removal part seem reasonable? I know that we're definitely not shipping PE but I don't know if that correlates to foss as well?

underscorgan:
@npwalker I hear you own platform lifecycle stuff, should PE + FOSS lose masters ~ the same time?

sarameisburger:
Alrighty updated to just remove for PE

sarameisburger:
I'll need to update the changelog too

PR #575, opened 2020-06-25
(EZ-35) Make use of project build-scripts

mwaggett:
This commit adds a method to copy build-scripts from the current project to the
staging dir. Previously, we only copied build-scripts from direct dependencies,
which resulted in the FOSS puppetserver build-scripts to be used instead of the
pe-puppet-server-extensions build-scripts when building from pe-pse instead of
from pe-puppetserver (PE-28810).

It appears that puppetserver gets around this by declaring a dependency on
itself in the :ezbake profile.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 This changelog adheres to [Keep a CHANGELOG](http://keepachangelog.com/).
 
 ## [Unreleased]
+Bugfix:
+  * (EZ-35) Make use of project build-scripts, in addition to dependency build-scripts.
 
 ## [2.1.7] - 2020-06-16
 Bugfix:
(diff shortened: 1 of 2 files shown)


mwaggett:
Tested this with various projects, both foss and PE, including my attempt to combine pe-pse and pe-puppetserver. 
Everything builds without issue, afaict.

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 This changelog adheres to [Keep a CHANGELOG](http://keepachangelog.com/).
 
 ## [Unreleased]
+Bugix:
```
mwaggett, line 6: wasn't really sure what to call this.. decided on bugfix, since I felt like I expected this behavior already but it wasn't there. ¯\_(ツ)_/¯

e-gris: 'Bugix:' --> 'Bugfix:'

mwaggett: lol whoops, thanks 😅

mwaggett: fixed

PR #578, opened 2020-08-17
(EZ-140) Allow excluding projects from ezbake resource search

Magisus:
Previously, ezbake would search all of the dependencies that it compiled
into the uberjar for ezbake config resources (e.g. CLI scripts and conf
files). However, we have a usecase now where one ezbake project is
wrapping another standalone ezbake project, so we want to exclude the
config from the wrapped project (puppetserver) and just use the fixtures
from the wrapping project (pe-puppetserver).

This commit adds a value, `excluded-ezbake-deps` to the `lein-ezbake`
config section, which contains a list of project symbols. These projects
will not be searched for packaging resources, but they will still be
included for compiling.

### DIFF
--- a/src/puppetlabs/ezbake/core.clj
+++ b/src/puppetlabs/ezbake/core.clj
@@ -454,9 +454,9 @@ Additional uberjar dependencies:
 
 (defn- prefix-project-name
   [project-name build-target]
-    (if (= build-target "pe")
-      (str "pe-" (name project-name))
-      (name project-name)))
+  (if (= build-target "pe")
+    (str "pe-" (name project-name))
+    (name project-name)))
 
 (defn generate-terminus-list
   [dependencies build-target]
@@ -634,7 +634,7 @@ Additional uberjar dependencies:
      :java-args                          (local->ruby :java-args
                                                       "-Xmx192m")
      :java-args-cli                      (local->ruby :java-args-cli "")
-     :tk-args                            (local->ruby :tk-args "" )
+     :tk-args                            (local->ruby :tk-args "")
      :bootstrap-source                   (-> (get-local :bootstrap-source :bootstrap-cfg)
                                              name as-ruby-literal)
      :logrotate-enabled                  (local->ruby :logrotate-enabled true)
@@ -838,7 +838,7 @@ Additional uberjar dependencies:
                                metadata (if (empty? munged-dependencies)
                                           {"version" version}
                                           {"version" version "dependencies" munged-dependencies})]]
-                         {(str project-name) metadata}))))
+                     {(str project-name) metadata}))))
 
 (defn generate-build-metadata-files
   [lein-project]
--- a/src/puppetlabs/ezbake/core.clj
+++ b/src/puppetlabs/ezbake/core.clj
@@ -880,6 +880,9 @@ Additional uberjar dependencies:
                            (deploy-snapshot lein-project)
                            (:version lein-project))
         reproducible? (not (System/getenv "EZBAKE_ALLOW_UNREPRODUCIBLE_BUILDS"))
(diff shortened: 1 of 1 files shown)


Magisus:
This still requires that you specify the current project as a dependency in the ezbake profile. Fixing that seemed like a bigger endeavor.

src/puppetlabs/ezbake/core.clj
```diff
@@ -880,6 +880,9 @@ Additional uberjar dependencies:
                            (deploy-snapshot lein-project)
                            (:version lein-project))
         reproducible? (not (System/getenv "EZBAKE_ALLOW_UNREPRODUCIBLE_BUILDS"))
+        ;; Do not copy ezbake config files and resources from the deps in this list.
```
mwaggett, line 883: Are these just the files in the `resources` directory (or I guess whatever the specified `:resource-paths` are)? If so, maybe worth mentioning that?

Magisus:
Example of this in action: https://github.com/puppetlabs/pe-puppet-server-extensions/pull/1174/commits/c0a4b7d40645d919ab9470ea43c4ae532fd7ab71

src/puppetlabs/ezbake/core.clj
```diff
@@ -880,6 +880,9 @@ Additional uberjar dependencies:
...
+        ;; Items should be symbols of the form "namespace/project"
+        excluded-ezbake-deps (get-in lein-project [:lein-ezbake :excluded-ezbake-deps])
```
mwaggett, line 885: hmm yeah coming up with a good name is tough.. I want something that is a little more specific about *what* is being excluded.. `:exclude-resources-from`? ¯\_(ツ)_/¯

src/puppetlabs/ezbake/core.clj
```diff
@@ -880,6 +880,9 @@ Additional uberjar dependencies:
                            (deploy-snapshot lein-project)
                            (:version lein-project))
         reproducible? (not (System/getenv "EZBAKE_ALLOW_UNREPRODUCIBLE_BUILDS"))
+        ;; Do not copy ezbake config files and resources from the deps in this list.
+        ;; Items should be symbols of the form "namespace/project"
```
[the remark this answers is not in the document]
justinstoller: No quotes, and iiuc, no leading colon unlike Molly's suggestion. Maybe ticks instead, like:
```
  :: Items should be symbols of the form `namespace/project`, e.g. `puppetlabs/puppetserver`
```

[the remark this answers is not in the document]
mwaggett: ah yeah, this is why I like examples 😅

src/puppetlabs/ezbake/core.clj
Magisus: Yeah I think that's what it wound up being. I wasn't sure if there was more stuff at first.

src/puppetlabs/ezbake/core.clj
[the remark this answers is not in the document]
Magisus: yes, like Justin said.

src/puppetlabs/ezbake/core.clj
Magisus: Yeah I can use that.

Magisus:
We currently don't generally search the current project's directories for basically anything. We still just look through the gathered dependencies. Fixing that would be a different set of changes, and even if we did fix that, we would probably also need something like this anyway, because we wouldn't want to make it ignore _all_ dependencies. There's still a valid workflow where an included project could bring in its own things (like a new config file or CLI script).

sarameisburger:
@Magisus can you add a short description about the changes/updates made under the `Unreleased` section of the changelog?

PR #579, opened 2020-08-17
(EZ-140) Update changelog for excluded resources setting

Magisus:


mwaggett:
We should probably just put this under a new version, if we're hoping to get this released immediately.

Magisus:
I was waiting to hear back from RE, but it sounds like they're down to cut it today. I'll update this to 2.2.0.

PR #583, opened 2020-12-01
(EZ-141) Install java 11 on EL >= 8

e-gris:

### DIFF
--- a/resources/puppetlabs/lein-ezbake/template/global/ext/fpm.rb
+++ b/resources/puppetlabs/lein-ezbake/template/global/ext/fpm.rb
@@ -194,10 +194,21 @@ if options.output_type == 'rpm'
   ##
   # Deal with Java issues. Install java 8 by default.
   # For rhel >= 8 we can use the newfangled 'or' syntax to allow for either
+
+  # Most yum-based platforms
   java_8  = 'java-1_8_0-openjdk-headless'
+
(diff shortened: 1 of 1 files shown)


e-gris:
Still needs testing.

e-gris:
Looks like fpm might not like the '(foo or bar)' syntax for --depend. Investigating.

underscorgan:
> Looks like fpm might not like the '(foo or bar)' syntax for --depend. Investigating.

aw dang

e-gris:
Progress. My local test almost works:

```
fpm -s dir --depends '(java-11-openjdk-headless or java-1.8.0-openjdk-headless)' -t rpm -n exgpkg source
```

However, the resulting package always prefers to install 1.8 regardless of the order specified. I'll keep poking.

underscorgan:
just to keep some discussion in gh for history, we were talking about this earlier and the package preferring 1.8 is weird, but ok, as long as if java-11 is installed it doesn't also try to install java-1.8

e-gris:
@Magisus PR lacks approval, likely because of my slightly disappointing and nebulously-phrased results. 

If it's OK to have java8 favored but java11 allowed, for some reason I've yet to uncover, give this an approval and I'll make the ezbake release.

e-gris:
@underscorgan We should merge or delete this. IIRC, you recently did some work with this.

PR #588, opened 2021-03-27
(REPLATS-150) Skip efforts to run as user when already service user

jpartlow:
Ran into this adjusting k8s pods to runas the service user. In this
case, the foreground command is already being executed as the service
user and attempts to become that user fail as it doesn't have privileges
to redundantly become itself anyway.


jpartlow:
@underscorgan should I refile this as an EZ issue or just maint it?

I don't know how notable it is :) but I can add it to the changelog if you like. It's a corner case I ran into.

underscorgan (approved):

underscorgan:
@jpartlow I don't think an EZ ticket is necessary, but a note in the changelog would be good!

underscorgan (approved):

PR #594, opened 2021-11-18
(maint) use rpm digest sha256

underscorgan:


e-gris (approved):
Just curious, why not sha512 ?

underscorgan:
sha256 is the current rpm default so seems 'safest'

PR #596, opened 2022-01-04
(RE-14469) Add debian 11 as a default FOSS platform

Magisus:
Debian 11 is not able to easily install Java 8, so like Debian 10 and
Ubunutu 20.04, we make packages depend on Java 11 instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Removed:
 
 Added:
   * Use sha256 digests for rpm packages.
+  * Add debian 11 to build targets
 
 Bugfix:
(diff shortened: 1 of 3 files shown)


Magisus:
Successfully tested this with Puppetserver, once I got an agent build with https://github.com/puppetlabs/puppet-agent/pull/2179

austb (approved):

mwaggett (approved):

Magisus:
I think the related agent PR is hung up on issues with vmpooler, but it doesn't block this. If we can get this in and released, that would be awesome.

Magisus:
Also, pretty sure the bit in the ticket about adding this to PE is a no-op here, since we don't build any master components for Debian in PE.

PR #601, opened 2022-02-12
Add ubuntu 20.04 and sles 15 as FOSS and PE master build targets

Magisus:


Magisus:
Did some minimal testing of this, more extensive testing needs to wait on agent updates and PE tarball support.

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ This changelog adheres to [Keep a CHANGELOG](http://keepachangelog.com/).
 
 ## [Unreleased]
 
+Added:
+  * (RE-14499) Add sles15 as a FOSS and PE master build target
+  * (RE-14541) Add ubuntu 20.04 as a FOSS and PE master build target
```
mwaggett, line 9:
```suggestion
  * (RE-14499) Add sles15 as a FOSS and PE primary server build target
  * (RE-14541) Add ubuntu 20.04 as a FOSS and PE primary server build target
```

resources/puppetlabs/lein-ezbake/template/pe/ext/build_defaults.yaml
```diff
@@ -1,12 +1,12 @@
 ---
 default_cow: 'base-bionic-amd64.cow'
-cows: 'base-bionic-amd64.cow'
+cows: 'base-bionic-amd64.cow base-focal-i386.cow'
```
mwaggett, line 3: huh, I wonder why bionic uses `amd64` here, but all the FOSS cows use `i386`..

isaac-hammes (approved):

resources/puppetlabs/lein-ezbake/template/pe/ext/build_defaults.yaml
Magisus: Oh, good catch, I think this should be amd64. I imagine it's because we wanted to preserve 32-bit support in FOSS but not in PE. But that's just speculation 🙂

mwaggett (approved):

PR #604, opened 2022-06-28
Migrate from travisci to github actions

underscorgan:


.github/workflows/pr.yml
```diff
@@ -0,0 +1,25 @@
...
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'temurin'
+          java-version: ${{ matrix.java }}
+      - name: setup lein
+        run: |
+          mkdir -p local/bin
+          curl --output local/bin/lein https://raw.githubusercontent.com/technomancy/leiningen/2.9.1/bin/lein
+          chmod +x local/bin/lein
+          sudo local/bin/lein
```
e-gris, line 24: Sort of a trivial point here but I'll ask anyway: is it possible to not use 'sudo'?

PR #607, opened 2022-10-26
(PE-34455) Add nrepl/nrepl to exclusion list

nmburgan:
https://github.com/puppetlabs/trapperkeeper/commit/f0c51cf3f8ce4e2d29f268ffed63b514348e43f9 added nrepl/nrepl to trapperkeeper, replacing org.clojure/tools.nrepl, but the ezbake exclusion was not updated, so nrepl is being included in built packages. While not enabled by default, this is a bit of a security risk, so we should exclude nrepl/nrepl as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 This changelog adheres to [Keep a CHANGELOG](http://keepachangelog.com/).
 
 ## [Unreleased]
+
+## [2.3.3] - 2022-10-26
+Maintenance:
+  * (PE-34484) Add nrepl/nrepl to exclusions list
+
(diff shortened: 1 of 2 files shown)


e-gris (approved):
Sigh, ezbake is nominally owned by RE but I've so little exposure/expertise with it, I feel sheepish approving PRs.

However, the changes seem match the intent, so yeah.

nmburgan:
This still isn't working properly, and despite our best efforts, we can't seem to effectively filter nrepl out here.  Going to take the easier route for now of just removing it from the AOT list for the individual components, rather than relying on ezbake to filter it out.

PR #611, opened 2023-04-04
Select java package version based on Puppet Platform Version

austb:
When not building a PE package, select the java dependency based on operating system and Puppet Platform version, which is assumed to be the major version of the package you are building.

This will allow us to maintain the same java versions between puppetserver and puppetdb, maintain a given dependency for an entire major version, and move that version forward between major versions.

- [x] Please add all notable changes to the "Unreleased" section of the CHANGELOG.

### DIFF
--- a/resources/puppetlabs/lein-ezbake/template/global/ext/fpm.rb
+++ b/resources/puppetlabs/lein-ezbake/template/global/ext/fpm.rb
@@ -184,7 +184,12 @@ if options.output_type == 'rpm'
       options.java =
         case options.platform_version
         when 8
-          'java-11-openjdk-headless'
+          # rpm on Redhat 7 may not support OR dependencies
+          if options.os_version == 7
+            'java-11-openjdk-headless'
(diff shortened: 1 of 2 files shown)


donoghuc:
I dont understand why we need to intuit this from the version and not explicitly set it as a configuration?

resources/puppetlabs/lein-ezbake/template/global/ext/fpm.rb
```diff
@@ -15,6 +15,7 @@
 options.sles = 0
 options.java = 'java-1.8.0-openjdk-headless'
 options.release = 1
+options.platform_version = 0
```
justinstoller, line 18: should this be `7`?

donoghuc (approved):
This looks extraordinarily complicated. If this will unblock us I think this is fine. Thanks for working through all this.

isaac-hammes (approved):

PR #620, opened 2023-08-03
(maint) Fix el9 platform 8 builds

austb:
Previously we were silently failing to specify a java dependency for Puppet Platform 8 builds on RedHat 9.

Add an error message that should help diagnose this issue for future redhat releases, and specify java 17 as a package dependency.


steveax (approved):
👍

PR #621, opened 2023-10-12
(PE-36861) Add a needrestart conf file for debian packages

nmburgan:
In Ubuntu 22, needrestart is now called by default whenever a package is installed via apt. Because our packages lay down a temp directory for use by the service, when the services create and modify files in those temp dirs, needrestart thinks the service needs to be restarted. Because we manage restarting our services as needed, this drops a file in /etc/needrestart/conf.d that adds the service to the list of services needrestart should ignore.

Note that this will drop the file regardless of if needrestart is present on the system (e.g. Ubuntu 18), but it should be harmless if needrestart isn't present.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 This changelog adheres to [Keep a CHANGELOG](http://keepachangelog.com/).
 
 ## [Unreleased]
+Added:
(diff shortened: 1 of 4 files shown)


e-gris (approved):
I'm not at all familiar with this part of ezbake but the PR looks quite sensible.

jonathannewman (approved):

PR #627, opened 2024-03-22
Use virtual Java packages on Red Hat and set java_bin

ekohl:
This uses the virtual packages jre-VERSION-headless instead of explicitly openjdk. This allows other JREs to provide the same.

It also uses an explicit path to the JRE specific java bin. At least on EL9 this allows the following upgrade path to work:

    dnf install https://yum.puppet.com/puppet7-release-el-9.noarch.rpm
    dnf install puppetserver
    dnf install https://yum.puppet.com/puppet8-release-el-9.noarch.rpm
    dnf upgrade puppetserver

Prior to this patch it would break, because it used /usr/bin/java which will point to Java 8. By using /usr/lib/jvm/jre-17/bin/java we know for sure it is Java 17.

### DIFF
--- a/resources/puppetlabs/lein-ezbake/template/foss/ext/default.erb
+++ b/resources/puppetlabs/lein-ezbake/template/foss/ext/default.erb
@@ -2,8 +2,8 @@
 # Init settings for <%= EZBake::Config[:project] %>
 ###########################################
 
-# Location of your Java binary (version 8)
-JAVA_BIN="/usr/bin/java"
+# Location of your Java binary
+JAVA_BIN="<%= EZBake::Config[:java_bin] %>"
 
 # Modify this if you'd like to change the memory allocation, enable JMX, etc
 JAVA_ARGS="<%= EZBake::Config[:java_args] %>"
--- a/resources/puppetlabs/lein-ezbake/template/foss/ext/default.erb
+++ b/resources/puppetlabs/lein-ezbake/template/foss/ext/default.erb
@@ -3,7 +3,7 @@
 ###########################################
 
 # Location of your Java binary
-JAVA_BIN="<%= EZBake::Config[:java_bin] %>"
+#JAVA_BIN=""
 
 # Modify this if you'd like to change the memory allocation, enable JMX, etc
 JAVA_ARGS="<%= EZBake::Config[:java_args] %>"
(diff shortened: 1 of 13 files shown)


resources/puppetlabs/lein-ezbake/template/global/ext/fpm.rb
```diff
@@ -181,25 +182,28 @@
...
+      when 8
+        # rpm on Redhat 7 may not support OR dependencies
+        if options.os_version == 7
+          options.java = 'jre-11-headless'
+          options.java_bin = '/usr/lib/jvm/jre-11/bin/java'
+        elsif options.os_version == 8
+          options.java = '(jre-17-headless or jre-11-headless)'
+          # TODO: which bin to use? /usr/bin/java may be anything
+        elsif options.os_version == 9
+          options.java = 'jre-17-headless'
```
ekohl, line 197: You can question why EL 9 always uses Java 17 while EL 8 uses 17 or 11. Both EL 8 and EL 9 provide both versions.

resources/puppetlabs/lein-ezbake/template/global/ext/fpm.rb
```diff
@@ -181,25 +182,28 @@
     options.systemd_el = 1
   elsif options.operating_system == :el && options.os_version >= 7 # systemd el
     if ! options.is_pe
+      # https://bugzilla.redhat.com/show_bug.cgi?id=2224427
```
ekohl, line 185: This bug is resolved, so it could be dropped. But this at least provides a reference of why it's there.

ekohl:
Now updated to also define `JAVA_BIN` in the systemd service file. See the additional commit for the reasoning.

resources/puppetlabs/lein-ezbake/template/foss/ext/default.erb
```diff
@@ -2,8 +2,8 @@
 # Init settings for <%= EZBake::Config[:project] %>
 ###########################################
 
-# Location of your Java binary (version 8)
-JAVA_BIN="/usr/bin/java"
+# Location of your Java binary
+#JAVA_BIN=""
```
ekohl, line 6: I found out that the `puppetserver` bin (which I can't find the source of now) also reads the defaults file. If that needs `JAVA_BIN` then this change is breaking.

ekohl:
Will open against https://github.com/OpenVoxProject/ezbake.

